PR #3, opened 2023-05-29
Native rumble & FF support

Rackover:
Proof of concept, kind of a hack, works for weapons as a demonstration but that's about it.
Proper implementation requires modifying & reallocating the infostring, which is kind of a pain. There was more info on the previous PR i opened about this but it got deleted unfortunately.
I need to clean up this branch at some point, but I don't know if i'll finish  this feature, so I'm leaving this here for reference :)

EDIT: I think it's done?


HeartbeatingForCenturies:
[UNRELATED] glad to see the project back in action

src/Components/Modules/Weapon.cpp
```diff
@@ -9,8 +9,86 @@ namespace Components
 
 	Game::WeaponCompleteDef* Weapon::LoadWeaponCompleteDef(const char* name)
 	{
+		std::string test = name;
+		if (test.ends_with("deserteaglegold_akimbo_mp"))
+		{
+			printf("");
+		}
+
```
Citizen01, line 17: That test was probably meant to be removed from the final PR

Rackover:
sue me

PR #40, opened 2023-08-28
imrpove logger code

diamante0018:


diamante0018:
on  a second thought i dont think this is necessary

PR #41, opened 2023-08-29
Update ZoneBuilder.cpp for building CoDOL Maps

maiyuzhe:
AssetHandler::ModifyAsset checks the zone version and then applies a patch, this won't work with maps compiled with zonebuilder because the ff version is the same as other iw4 zones.  Therefore the sortkey needs to be fixed when the map gets built instead of at runtime.

### DIFF
--- a/src/Components/Modules/ZoneBuilder.cpp
+++ b/src/Components/Modules/ZoneBuilder.cpp
@@ -259,29 +259,6 @@ namespace Components
 			return false;
 		}
 
-		//patch for codol maps dumped with zonetool, the FF version is changed, so the sortKey hack in AssetHandler::ModifyAsset won't work
-		if (type == Game::XAssetType::ASSET_TYPE_GFXWORLD)
-		{
-			if (assetHeader.gfxWorld->sortKeyDistortion == 44)
-			{
-				assetHeader.gfxWorld->sortKeyDistortion = 43;
-			}
-			//Absolutely necessary patch for later codol maps. prevents player models from glowing ridiculously 
-			if (assetHeader.gfxWorld->heroOnlyLightCount > 0)
-			{
-				Logger::Print("Fixing heroOnlyLightCount...\n");
-				assetHeader.gfxWorld->heroOnlyLightCount = 0;
-			}
-		}
-		//likewise if material's sortKey is 44, it needs to be changed to 43.
-		if (type == Game::XAssetType::ASSET_TYPE_MATERIAL)
-		{
-			if (assetHeader.material->info.sortKey == 44)
-			{
-				assetHeader.material->info.sortKey = 43;
-			}
-		}
-
 		Game::XAsset asset;
 		asset.type = type;
 		asset.header = assetHeader;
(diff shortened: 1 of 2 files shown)


Rackover:
This change will execute on every single map, including native IW4 maps, ported IW3 and IW5 maps - in the vast majority, maps that don't need that patch. You are right that the CODO conversion code is incomplete and/or improper, but adding a second conversion code snippet into ModifyAsset that executes for every single map is not the way to go.

If you want to fix that issue, you need to check Zones.cpp and apply those changes once the asset has been read/converted on the fly. You might want to check the functions for GfxWorld and Material asset types in that file, and apply your changes there, so that only a targeted CODO benefits from those changes instead of blanket bombing the entire game with ModifyAsset 🙂

src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -259,6 +259,29 @@ namespace Components
 			return false;
 		}
 
+		//patch for codol maps dumped with zonetool, the FF version is changed, so the sortKey hack in AssetHandler::ModifyAsset won't work
+		if (type == Game::XAssetType::ASSET_TYPE_GFXWORLD)
```
Rackover, line 263: Will apply to every zone - not a good patch, and in fact, will introduce several problems.
Some maps might need hero light, and some vailla materials might use sort key 44 for something else than distortion.
If you have CODO compatibility patches to submit, you need to edit the appropriate functions in Zones.cpp so they only apply to a specific CODO version rather than to every single asset ever built.

As a general rule - **no conversion of any kind should ever happen when building**. The files on disk should be built into the zone, **identical**.  Conversion should only happen in Zones.cpp or, preferably, in an external program (iw3xport/iw5xport/etc)

Rackover (changes requested):

Rackover:
Here's a few pointers: take it with a grain of salt cause i have not checked very thoroughly.

You can set the sort key/herolight of GFXWorld here : https://github.com/iw4x/iw4x-client/blob/r4380/src/Components/Modules/Zones.cpp#L2008 by `reinterpret_cast` ing `buffer` as `Game::GfxWorld*`

You can modify the material either after this line:
https://github.com/iw4x/iw4x-client/blob/r4380/src/Components/Modules/Zones.cpp#L1934
Or around here https://github.com/iw4x/iw4x-client/blob/r4380/src/Components/Modules/Zones.cpp#L3421 but don't forget to check the zone version.

maiyuzhe:
Implemented the fixes you suggested, I decided to just let heroOnlyLights be, because it seems only mp_abandon_sh_mk seems to have that issue.  Please let me know if there any further issues.

Rackover:
Please test that the current maps (nuked, bloc, etc ...) work without issues with your patch and report back to me. If everything is good this can be merged 🙂

maiyuzhe:
> Please test that the current maps (nuked, bloc, etc ...) work without issues with your patch and report back to me. If everything is good this can be merged 🙂

No issues with the  IW4x DLC on my end.

Rackover (approved):

PR #68, opened 2023-12-05
Compatibility and improvements for Bot Warfare

ineedbots:
**What does this PR do?**

My Bot Warfare mod (https://github.com/ineedbots/iw4_bot_warfare) will be compatible with this client with these changes. Bots will be able to press the Use button on objects, bots will be able to use alt weaponry properly. Added `botMeleeParams` builtin to allow bots to melee lung

Nothing is breaking.

Max SL string length is actually 65535, not 1023, this is a required change for Bot Warfare reading waypoints. "scriptdata/" is NOT added to filepath if it already exists in the argument.

- [x] Mention any [related issues](https://github.com/XLabsProject/iw4x-client/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [x] Follow our [coding conventions](https://github.com/XLabsProject/iw4x-client/blob/master/CODESTYLE.md)

### DIFF
--- a/src/Components/Modules/Bots.cpp
+++ b/src/Components/Modules/Bots.cpp
@@ -301,8 +301,10 @@ namespace Components
 			return;
 		}
 
+		auto clientNum = cl - Game::svs_clients;
+
 		// Keep test client functionality
-		if (!g_botai[cl - Game::svs_clients].active)
+		if (!g_botai[clientNum].active)
 		{
 			Game::SV_BotUserMove(cl);
 			return;
@@ -313,11 +315,11 @@ namespace Components
 
 		userCmd.serverTime = *Game::svs_time;
 
-		userCmd.buttons = g_botai[cl - Game::svs_clients].buttons;
-		userCmd.forwardmove = g_botai[cl - Game::svs_clients].forward;
-		userCmd.rightmove = g_botai[cl - Game::svs_clients].right;
-		userCmd.weapon = g_botai[cl - Game::svs_clients].weapon;
-		userCmd.primaryWeaponForAltMode = g_botai[cl - Game::svs_clients].lastAltWeapon;
+		userCmd.buttons = g_botai[clientNum].buttons;
+		userCmd.forwardmove = g_botai[clientNum].forward;
+		userCmd.rightmove = g_botai[clientNum].right;
+		userCmd.weapon = g_botai[clientNum].weapon;
+		userCmd.primaryWeaponForAltMode = g_botai[clientNum].lastAltWeapon;
 
 		userCmd.angles[0] = ANGLE2SHORT((cl->gentity->client->ps.viewangles[0] - cl->gentity->client->ps.delta_angles[0]));
 		userCmd.angles[1] = ANGLE2SHORT((cl->gentity->client->ps.viewangles[1] - cl->gentity->client->ps.delta_angles[1]));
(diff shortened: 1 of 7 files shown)


src/Components/Modules/GSC/IO.cpp
```diff
@@ -85,7 +93,7 @@ namespace Components::GSC
 			return;
 		}
 
-		char line[1024]{};
+		char line[65536]{};
```
diamante0018, line 96: This change should be undone
There is no sane person that goes to the great length to read a file in bits as a stream than just uses such a big buffer, ON THE STACK???!!!

diamante0018 (changes requested):

src/Components/Modules/GSC/IO.cpp
diamante0018: Resolved

src/Components/Modules/GSC/IO.cpp
```diff
@@ -30,12 +30,20 @@ namespace Components::GSC
 	std::filesystem::path IO::BuildPath(const char* path)
 	{
 		const std::filesystem::path fsGame = (*Game::fs_gameDirVar)->current.string;
+
+		// check if we need to append scriptdata folder, for backwards compat
+		std::string spath = path;
+		if (!spath.starts_with("scriptdata/") && !spath.starts_with("scriptdata\\"))
+		{
+			spath = "scriptdata/" + spath;
```
Rackover, line 38: Maybe turning `scriptdata` into a constant would be nice since it appears occasionally throughout this file, rather than repeating the string

src/Components/Modules/Bots.cpp
```diff
@@ -316,6 +317,7 @@ namespace Components
 		userCmd.forwardmove = g_botai[cl - Game::svs_clients].forward;
 		userCmd.rightmove = g_botai[cl - Game::svs_clients].right;
 		userCmd.weapon = g_botai[cl - Game::svs_clients].weapon;
+		userCmd.primaryWeaponForAltMode = g_botai[cl - Game::svs_clients].lastAltWeapon;
```
Rackover, line 320: i skimmed through the function and you use `cl - Game::svs_clients` multiple times, same thing I think it would be nice if it was factorized into a single variable at the beginning of your function

src/Components/Modules/Bots.cpp
```diff
@@ -339,11 +341,38 @@ namespace Components
...
 		{
 			g_botai[clientNum].weapon = static_cast<uint16_t>(iWeaponIndex);
+			g_botai[clientNum].lastAltWeapon = 0;
+
+			auto* def = Game::BG_GetWeaponCompleteDef(iWeaponIndex);
+
+			if (def->weapDef->inventoryType == Game::WEAPINVENTORY_ALTMODE)
+			{
+				auto* ps = &Game::g_entities[clientNum].client->ps;
+				auto num_weaps = Game::BG_GetNumWeapons();
```
Rackover, line 356: sir this is camelCase territory

src/Components/Modules/Bots.cpp
```diff
@@ -472,6 +501,11 @@ namespace Components
 		});
 	}
 
+	bool Bots::Player_UpdateActivate_stub(int)
+	{
+		return false;
```
Rackover, line 506: Does this have no other consequences?  What was the original gentity check at 0x4B3D00 done for? 🤔

Rackover (commented):
Looks good and I will test it, if it runs i see no problem merging that after the usual nitpicky formatting has been taken care of 😗

src/Components/Modules/Bots.cpp
Rackover: thx!! if you can add a comment in the code saying that then it's nice.
Do you know if this could mess up with testclients (non-bw bots) ?

Rackover (approved):

PR #70, opened 2023-12-15
fix: auto-restarting the same map in private match

WantedDV:
Fixes issue where private matches would map_restart after every game. Players are returned to the lobby instead.

Applies to private matches only.


wroyca:
Closing since we don't have a lobby in a private match to begin with, but do reopen if there's a misunderstanding here. :saluting_face:

wroyca (approved):

PR #76, opened 2023-12-29
feature: auto updater!

diamante0018:
**What does this PR do?**

- Auto updater
- Fixes singleton component memory leak

### DIFF
--- a/src/Components/Modules/Updater.cpp
+++ b/src/Components/Modules/Updater.cpp
@@ -41,14 +41,14 @@ namespace Components
 			if (!parseResult || !doc.IsObject())
 			{
 				// Nothing to do in this situation. We won't know if we need to update or not
-				Logger::Print("GitHub sent an invalid reply\n");
+				Logger::Print("GitHub sent an invalid reply (malformed JSON)\n");
 				return;
 			}
 
 			if (!doc.HasMember("tag_name") || !doc["tag_name"].IsString())
 			{
 				// Nothing to do in this situation. We won't know if we need to update or not
-				Logger::Print("GitHub sent an invalid reply\n");
+				Logger::Print("GitHub sent an invalid reply (missing 'tag_name' JSON member)\n");
 				return;
 			}
 
--- a/src/Components/Modules/Updater.cpp
+++ b/src/Components/Modules/Updater.cpp
@@ -6,8 +6,6 @@
 #include <Utils/WebIO.hpp>
 
 #include <rapidjson/document.h>
-#include <rapidjson/prettywriter.h>
-#include <rapidjson/stringbuffer.h>
 
 namespace Components
 {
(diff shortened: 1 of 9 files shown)


src/Components/Modules/Updater.cpp
```diff
@@ -0,0 +1,106 @@
...
+		UIScript::Add("getAutoUpdate", [](const UIScript::Token& /*token*/, const Game::uiInfo_s* /*info*/)
+		{
+			if (const auto exe = GetLauncher(); exe.has_value())
+			{
+				Game::Sys_QuitAndStartProcess(exe.value().data());
+				return;
+			}
+
+			// No launcher was found on the system, time to tell them to download it from GitHub
+			Utils::OpenUrl("https://forum.alterware.dev/t/how-to-install-the-alterware-launcher/56");
```
Rackover, line 103: Sorry to be extra annoying but i would like to ask you to please turn all those addresses and filename (http://forum, alterware-launcher.exe, the github repo URL, etc...) into variables or macros at the beginning of the file or in the headerfile, however you like it. thanks! 

Other than that it's good to go, so whenever it's done i will test it and merge it.
Does this have a dependency on any IW4x rawfile PR you would want this to get tested with? So I can test both together & greenlight both!

src/Components/Modules/Updater.cpp
```diff
@@ -0,0 +1,106 @@
...
+		// If they use the alterware-launcher once to install they will have this file
+		// If they don't, what are they waiting for?
+		constexpr auto* revisionFile = ".iw4xrevision";
+
+		void CheckForUpdate()
+		{
+			std::string revision;
+			if (!Utils::IO::ReadFile(revisionFile, &revision) || revision.empty())
+			{
+				Logger::Print(".iw4xrevision does not exist. Notifying the user an update is available\n");
```
Rackover, line 27: missed an occasion to std::format here
`Logger::Print("{} does not exist", revisionFile)`

Rackover (changes requested):

src/Components/Modules/Updater.cpp
diamante0018: Yeah both should be merged at the same time

src/Components/Modules/Updater.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+		// If they use the alterware-launcher once to install they will have this file
+		// If they don't, what are they waiting for?
+		constexpr auto* REVISION_FILE = ".iw4xrevision";
+		constexpr auto* GITHUB_REMOTE_URL = "https://api.github.com/repos/iw4x/iw4x-client/releases/latest";
+		constexpr auto* INSTALL_GUIDE_REMOTE_URL = "https://forum.alterware.dev/t/how-to-install-the-alterware-launcher/56";
+
+		void CheckForUpdate()
+		{
+			std::string revision;
```
Rackover, line 24: no `{}` ?

src/Components/Modules/Updater.cpp
```diff
@@ -0,0 +1,107 @@
...
+			{
+				// Nothing to do in this situation. We won't know if we need to update or not
+				Logger::Print("GitHub sent an invalid reply\n");
+				return;
+			}
+
+			if (!doc.HasMember("tag_name") || !doc["tag_name"].IsString())
+			{
+				// Nothing to do in this situation. We won't know if we need to update or not
+				Logger::Print("GitHub sent an invalid reply\n");
```
Rackover, line 51: Its preferable not to have two different errors print out the same message. You can add (malformed JSON) on the first one and (missing tag_name) on the second for instance.

Rackover (changes requested):
Two small things to address and then i'll merge it with my eyes closed

src/Components/Modules/Updater.cpp
diamante0018: That has no functional change: {} -> calls the default ctor which is the same as my code.

Rackover (approved):

PR #78, opened 2024-01-11
Feat: Native integration with Fail2Ban for Linux server Hosters

diamante0018:
**What does this PR do?**

Allows the user to harden their servers from unwanted attacks such as RCon password brute force.
This PR adds a few prints to critical sections of the code where I think a possible attacker may attempt to send invalid data.

- Adds a dvar that controls where the log file is written default: `/var/iw4x.log` (taken from fail2ban docs. For testing I used to change it to `userraw/logs/iw4x.log`. 
- Adds command line arg to set this feature one.

I used these fail2ban configs and I tested them well enough confirming they work out of the box https://git.alterware.dev/AlterWare/iw4x-fail2ban

Once this PR merged I encourage the creation of a similar repo or otherwise distribute the files under this github org

- [X ] Focus on a single fix or feature; remove any unrelated formatting or code changes

### DIFF
--- a/src/Components/Modules/Auth.cpp
+++ b/src/Components/Modules/Auth.cpp
@@ -144,6 +144,7 @@ namespace Components
 		Proto::Auth::Connect connectData;
 		if (msg->cursize <= 12 || !connectData.ParseFromString(std::string(reinterpret_cast<char*>(&msg->data[12]), msg->cursize - 12)))
 		{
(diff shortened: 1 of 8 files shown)


Rackover:
Looking at the diff this seems to have a dependency on https://github.com/iw4x/iw4x-client/pull/76 so i'll wait to have 76 in, then you can rebase and merge this as an independant feature, looks good otherwise

diamante0018:
Yeah.. My bad.

Rackover (approved):

PR #79, opened 2024-01-13
fix(download): restore password verification

diamante0018:
Old feature by momo, his old code didn't work because he forgot to to call dump hex on the server password.
I removed it during the great refactoring of download.cpp when we upgraded past 4 major releases of mongoose.

Now I add it back

### DIFF
--- a/src/Components/Modules/Download.cpp
+++ b/src/Components/Modules/Download.cpp
@@ -478,14 +478,14 @@ namespace Components
 
 		if (len <= 0)
 		{
-			ReplyError(c, 403, "Password Required");
+			Download::ReplyError(c, 403, "Password Required");
 			return false;
 		}
 
 		const auto password = std::string(buffer, len);
 		if (password != Utils::String::DumpHex(Utils::Cryptography::SHA256::Compute(g_password), ""))
 		{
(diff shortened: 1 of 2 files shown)


src/Components/Modules/Download.cpp
```diff
@@ -462,6 +462,40 @@ namespace Components
...
+	bool VerifyPassword([[maybe_unused]] mg_connection* c, [[maybe_unused]] const mg_http_message* hm)
+	{
+		const std::string g_password = *Game::g_password ? (*Game::g_password)->current.string : "";
+		if (g_password.empty()) return true;
+
+		// SHA256 hashes are 64 characters long but we're gonna be safe here
+		char buffer[128]{};
+		const auto len = mg_http_get_var(&hm->query, "password", buffer, sizeof(buffer));
+
+		const auto reply = [&c](const char* s) -> void
```
Rackover, line 474: Instead of rewritting it here as a local, could you please use `ReplyError()` function instead, and modify `ReplyError` to accept a custom optional message if you want to send a specific string?

Rackover (changes requested):

Rackover (approved):

PR #83, opened 2024-01-20
feat: containers!

diamante0018:
Tested by me and xerxes, see discord of images of iw4x running in eggs

### DIFF
--- a/src/Components/Modules/FastFiles.cpp
+++ b/src/Components/Modules/FastFiles.cpp
@@ -279,6 +279,7 @@ namespace Components
 
 		for (auto& path : paths)
 		{
+			const auto* dir = (*Game::fs_basepath)->current.string;
 			auto absoluteFile = std::format("{}\\{}{}", dir, path, file);
 
 			// No ".ff" appended, append it manually
--- a/src/Components/Modules/FastFiles.cpp
+++ b/src/Components/Modules/FastFiles.cpp
@@ -524,6 +524,27 @@ namespace Components
 		}
 	}
 
+	bool FastFiles::DB_FileExists_Hk(const char* zoneName, Game::FF_DIR source)
+	{
+		char filename[256]{};
+
+		DB_BuildOSPath_FromSource_Default(zoneName, source, sizeof(filename), filename);
+		if (auto zoneFile = Game::Sys_OpenFileReliable(filename); zoneFile != INVALID_HANDLE_VALUE)
+		{
+			CloseHandle(zoneFile);
+			return true;
+		}
+
+		DB_BuildOSPath_FromSource_Custom(zoneName, source, sizeof(filename), filename);
+		if (auto zoneFile = Game::Sys_OpenFileReliable(filename); zoneFile != INVALID_HANDLE_VALUE)
+		{
+			CloseHandle(zoneFile);
+			return true;
+		}
+
+		return false;
+	}
+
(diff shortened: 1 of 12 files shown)


diamante0018:
@Rackover simple docs on how to use this have been posted and xerxes helped review the overall general view of the feature.

I'm not a server hoster so I can't actually provide pre-made docker files although I did my best to explain what needs to be done for it to happen.

If someone wants to write docker it would be pretty cool

https://forum.alterware.dev/t/docker-containers-for-iw4x-linux/316

src/Components/Modules/FastFiles.cpp
```diff
@@ -506,11 +506,74 @@ namespace Components
...
+			(void)sprintf_s(filename, size, "%s\\%s%s.ff", FileSystem::Sys_DefaultInstallPath_Hk(), GetZoneLocation(zoneName), zoneName);
+		}
+	}
+
+	void FastFiles::DB_BuildOSPath_FromSource_Custom(const char* zoneName, Game::FF_DIR source, unsigned int size, char* filename)
+	{
+		// TODO: this is where user map and mod.ff check should happen
+		if (source == Game::FFD_DEFAULT)
+		{
+			(void)sprintf_s(filename, size, "%s\\%s%s.ff", FileSystem::Sys_HomePath_Hk(), GetZoneLocation(zoneName), zoneName);
```
Rackover, line 523: (void) ?

src/Utils/Utils.cpp
```diff
@@ -116,26 +117,54 @@ namespace Utils
...
 		char* buffer{};
 		std::size_t size{};
-		if (_dupenv_s(&buffer, &size, "MW2_INSTALL") != 0 || buffer == nullptr)
+		if (_dupenv_s(&buffer, &size, "IW4x_INSTALL") != 0 || buffer == nullptr)
 		{
 			SetLegacyEnvironment();
 			return;
 		}
 
-		const auto _0 = gsl::finally([&] { std::free(buffer); });
```
Rackover, line 129: why was this here?
I find it in other places of the codebase. Using gsl::finally to free instead of just freeing.

Rackover (commented):
Look OK but I need to check those addresses when I have time

src/Components/Modules/FastFiles.cpp
diamante0018: It's a trait I have acquired because it's very good practice to use void when you are intentionally discarding the return value of s function.
sprintf returns the number of written characters which we don't need in this case.
If I don't use void it's up to the reviews to guess if I didn't use the return value because I forgot it or it was intentional.

src/Utils/Utils.cpp
diamante0018: I could put back gsl free however I found that sometimes it works unexpectedly by freeing the value (calling the fctor that is the lamba function) before the end of the function so after they I kinda got spooked and I'm not sure when I'll be using it again

Rackover: ok i'm really confused about this. I think this should stay because it's always been there, and so far so good.
But i don't have many more arguments to this. If you think this is dangerous and can make the game unstable we remove it. But i would rather advance with certainty. tell me what you think we should do

diamante0018: done

Rackover (approved):

PR #86, opened 2024-01-28
🦴

Rackover:
🦴🦴🦴

### DIFF
--- a/src/Components/Modules/AssetInterfaces/IXModel.cpp
+++ b/src/Components/Modules/AssetInterfaces/IXModel.cpp
@@ -11,8 +11,19 @@ namespace Assets
 	{
 		header->model = builder->getIW4OfApi()->read<Game::XModel>(Game::XAssetType::ASSET_TYPE_XMODEL, name);
 
+		if (!header->model)
+		{
+			// In that case if we want to convert it later potentially we have to grab it now:
+			header->model = Game::DB_FindXAssetHeader(Game::XAssetType::ASSET_TYPE_XMODEL, name.data()).model;
+		}
+
 		if (header->model)
 		{
+			if (Components::ZoneBuilder::zb_sp_to_mp.get<bool>())
+			{
+				Assets::IXModel::ConvertPlayerModelFromSingleplayerToMultiplayer(header->model, *builder->getAllocator());
+			}
+
 			// ???
 			if (header->model->physCollmap)
 			{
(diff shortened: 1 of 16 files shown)


src/Components/Modules/AssetInterfaces/IXModel.cpp
```diff
@@ -1,14 +1,29 @@
 #include <STDInclude.hpp>
+
+#include <stdint.h>
```
diamante0018, line 3: that's strange, are you sure you need this? I would imagine there is a cpp equivalent already in STDInclude

src/Components/Modules/AssetInterfaces/IXModel.cpp
```diff
@@ -257,4 +272,733 @@ namespace Assets
...
+			std::memcpy(&newQuats[(atPositionM1 + 1) * 4], &model->quats[atPositionM1 * 4], sizeof(uint16_t) * 4 * lengthOfSecondPartM1);
+			std::memcpy(&newTrans[(atPositionM1 + 1) * 3], &model->trans[atPositionM1 * 3], sizeof(float) * 3 * lengthOfSecondPartM1);
+		}
+
+		//Game::Z_VirtualFree(model->baseMat);
+		//Game::Z_VirtualFree(model->boneInfo);
+		//Game::Z_VirtualFree(model->boneNames);
+		//Game::Z_VirtualFree(model->quats);
+		//Game::Z_VirtualFree(model->trans);
+		//Game::Z_VirtualFree(model->parentList);
```
diamante0018, line 594: Yeah I was about to say since this is zonebuilder who cares if it leaks a bit of memory.
The OS will clean up the memory after zb exits

src/Components/Modules/ZoneBuilder.cpp
```diff
@@ -1389,6 +1476,24 @@ namespace Components
...
 				});
 
+			Command::Add("buildmod", [](const Command::Params* params)
+				{
+					if (params->size() < 2) return;
+
+					std::string modName = params->get(1);
+					Logger::Print("Building zone for mod '{}'...\n", modName);
+
+					const std::string previousFsGame = Dvar::Var("fs_game").get<std::string>();
```
diamante0018, line 1486: i feel like this code could use a bit of rewriting. consider storing Dvar::Var("fs_game") in a local variable once instead of looking for it multiple times.

diamante0018 (commented):
I am not an expert on Zone code so I can't say much about it besides that I didn't see anything that needs addressing.
There seems to be in the diff file a few asm stubs that had their code style severely borked so it might be hard to read those later, consider fixing them manually.
✨✨

src/Components/Modules/AssetInterfaces/IXModel.cpp
Rackover: I wish I could free those but no matter what i try it halts and catches fire, so, uuh, freeing memory will be left as an exercise for the reader

src/Components/Modules/AssetInterfaces/IXModel.cpp
Rackover: it appeared while i wasn't looking

src/Components/Modules/AssetInterfaces/IXModel.cpp
diamante0018: There is a chance the entire memory allocation is done with virtual alloc once somewhere with mem_reserve flag 
which is the "committed" (winapi flag) later

So this might not leak **IF** at the end the big chunk that was reserved is freed with the right flags

diamante0018 (approved):
i think this looks good.

Issue #72, opened 2023-12-19
Chinatown crashes on startup, sometimes


Rackover:
Game crashes with the error
`No game type scripts found in maps/mp/gametypes folder`

Happens inconsistently
repro steps 2/3 : 
- open main menu
- developer 2
- developer_script 1
- private match, select chinatown, select domination, start game

Rackover:
Cannot reproduce on a clean install

PR #91, opened 2024-02-16
fix: removed hook that can crash when loading zone

WantedDV:
**What does this PR do?**

Removing this hook fixes the crash when loading into Chinatown. 
Closes https://github.com/iw4x/iw4x-client/issues/72

I did not pick up on any unwanted changes.

It might be worth looking into the Chinatown to figure out what sets it apart from the rest. One thing I noted was that the loading screen for this map seems to struggle with showing the correct map name: it will start off with "Chinatown" briefly, then show "Unknown", and finally, "Chinatown" again. The console reports `Overriding asset 'MPUI_CARENTAN' of type localize`. This is the only map I noticed that would do this.

### DIFF
--- a/src/Components/Modules/Maps.cpp
+++ b/src/Components/Modules/Maps.cpp
@@ -847,11 +847,9 @@ namespace Components
 
 		// Always refresh arena when loading or unloading a zone
 		Utils::Hook::Nop(0x485017, 2);
-		Utils::Hook::Nop(0x4FD8C7, 2); // Gametypes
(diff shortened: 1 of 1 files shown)


Rackover (approved):
Sorry for the delay, don't hesitate to ping me if nobody shows up to review your PR after a week or two :)
I've tested this change on a branch and it works

Regarding Chinatown, I don't know exactly what's up with the issues you encounter. I know the map has a few oddities, in particular its GSC is a bit messy, but nothing that should trigger the things you mention - so I'm not sure.

PR #95, opened 2024-02-27
feat: added -lan flag

WantedDV:
**What does this PR do?**

Adds `-lan` startup flag for offline/lan gaming. 

Disables node system which would print nearly 600 error messages to the console as a result of Sys_SendPacket failing when offline (called from `Node::Entry::sendRequest`).

Master server still works should the user regain internet connection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,11 +33,12 @@
 | `-nointro`              | Skip game's cinematic intro.                   |
 | `-version`              | Print IW4x build info on startup.              |
 | `-nosteam`              | Disable friends feature and do not update Steam about the game's current status just like an invisible mode. |
+| `-lan`                  | Disable node system for the client/server and disable sending heartbeats to the master server. |
 | `-unprotect-dvars`      | Allow the server to modify saved/archive dvars. |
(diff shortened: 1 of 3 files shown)


diamante0018:
missing documentation of this flag in README.md

Rackover:
This new flag seems entirely equivalent to `sv_lanOnly`
In fact i feel like you can achieve the exact same behaviour by just adding `+set sv_lanOnly 1`  to the game launch options, which will act just the same way as your new flag except it will not require any new implementation. 

On the basis that this feature already exists, I think this PR does not warrant merging?

Noobgamer0111 (approved):

Rackover:
Looking at your implementation, everywhere where you use your flag you could also just use SVLanOnly
The Dvar works whether you're a dedicated server or not, but maybe it indeed doesn't need to be in Dedicated.cpp and could belong in another component. This does not warrant the addition of a new flag, imo
```
- if (Dedicated::IsEnabled() && Dedicated::SVLanOnly.get<bool>()) return;
+ if (Network::SVLanOnly.get<bool>()) return;
```
for instance?

PR #106, opened 2024-03-30
RCon: remove ECC & related functions + fix logic bug

diamante0018:
**What does this PR do?**

- Remove momo5502's funny code
- Add a return statement when RSA key (needed for safe RCon patch) is invalid

- Make the client not crash when an invalid RSA key is detected
- Make client better

No

- [X] Mention any [related issues](https://github.com/XLabsProject/iw4x-client/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [X] Follow our [coding conventions](https://github.com/XLabsProject/iw4x-client/blob/master/CODESTYLE.md)

### DIFF
--- a/src/Components/Modules/RCon.cpp
+++ b/src/Components/Modules/RCon.cpp
@@ -11,9 +11,6 @@ namespace Components
 
 	std::vector<std::size_t> RCon::RConAddresses;
 
-	RCon::Container RCon::RConContainer;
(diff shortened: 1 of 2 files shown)


Rackover (approved):
Looks good and @diamante0018  said they tested it, so it's good to go as far as I'm concerned.

PR #107, opened 2024-03-30
fix command line arguments for alterware-launcher.exe

mxve:
bif gix

### DIFF
--- a/src/Utils/Library.cpp
+++ b/src/Utils/Library.cpp
@@ -102,22 +102,22 @@ namespace Utils
 		this->module_ = nullptr;
 	}
 
-	void Library::LaunchProcess(const std::string& process, const std::string& commandLine, const std::string& currentDir)
+	void Library::LaunchProcess(const std::wstring& process, const std::wstring& commandLine, const std::filesystem::path& currentDir)
 	{
-		STARTUPINFOA startupInfo;
-		PROCESS_INFORMATION processInfo;
+		STARTUPINFOW startup_info;
+		PROCESS_INFORMATION process_info;
 
-		ZeroMemory(&startupInfo, sizeof(startupInfo));
-		ZeroMemory(&processInfo, sizeof(processInfo));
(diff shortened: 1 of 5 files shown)


diamante0018 (approved):
We tested this locally and this fixes an issue with the alterware launcher. please merge

mxve:
probably still requires some work, don't merge yet

diamante0018 (approved):
Very good

Rackover (commented):
It is not immediatly obvious what this fixes, can I have more info please? 😂 
Was this tested on a wide string locale system, like Russian, to make sure there's no regression?  (Since we're touching environment variables and system path name)

diamante0018:
This PR was done to fix IW4x auto restarting via the alterware launcher post self-update.
It was tested on mxve's PC and mine. Yes this will absolutely not break anything on russian PC because we are using widestrings where possible which can fit non utc-8 chars like russian or japanese.

src/Utils/Library.cpp
```diff
@@ -102,22 +102,22 @@ namespace Utils
...
+			nullptr, false, NULL, nullptr, currentDir.wstring().data(),
+			&startup_info, &process_info);
 
-		if (processInfo.hThread && processInfo.hThread != INVALID_HANDLE_VALUE)
-			CloseHandle(processInfo.hThread);
-		if (processInfo.hProcess && processInfo.hProcess != INVALID_HANDLE_VALUE)
-			CloseHandle(processInfo.hProcess);
+		if (process_info.hThread && process_info.hThread != INVALID_HANDLE_VALUE)
+			CloseHandle(process_info.hThread);
+		if (process_info.hProcess && process_info.hProcess != INVALID_HANDLE_VALUE)
```
Rackover, line 120: It should be camelCase not snake_case, in IW4x that's what is used

Rackover (approved):
Looks good just a little formatting issue. I will merge as soon as it is addressed, and i will tag a new release

PR #111, opened 2024-04-23
Optimizing the game's Huffman (de)compression code

Caball009:
**What does this PR do?**
1. This pr adds a more optimized version of the game's huffman functions for (de)compression of data, particularly the compression. The game mainly uses huffman coding to compress game data on the server side and to decompress it on the client side.
2. If so desired, but not currently implemented, this may be a better way to fix the exploit issues related to huffman decompression. It's certainly a more efficient way than using a static buffer and copying the data if there are no buffer overruns as is currently done in the 'Security' namespace. https://github.com/iw4x/iw4x-client/blob/develop/src/Components/Modules/Security.cpp#L6-L30

This pr should not change existing behavior in any way.

1. There's currently a potential for OOB read for decompression in the game's huffman code. The maximum number of OOB bytes is 2 bytes. The main reason to fix this would probably be correctness and to be compliant with sanitizers, not necessarily security. This is currently not fixed in this pr, but can be if so desired. 
2. The game's functions 'MSG_ReadBitsCompress' and 'MSG_WriteBitsCompress' are currently not hooked and replaced with the functions in this pr. I'd like to get some feedback before proceeding to do so.
That would also mean the game's huffman initialization can be skipped, although this only costs a couple of ms in my experience.
3. If the 'IterateTree' function is not inlined by the compiler, it'd probably be better to make it a part of the 'Decompress' function if optimization is the goal.

- [x] Mention any [related issues](https://github.com/XLabsProject/iw4x-client/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [x] Follow our [coding conventions](https://github.com/XLabsProject/iw4x-client/blob/master/CODESTYLE.md)


Caball009:
@Rackover (you told me to ping you).

Rackover:
Thanks for pinging me, I'm not good enough with Huffman to assert of the quality of this implementation, however I see that you did similar work before and I trust that you know what you've doing.

I think I would like to "test" the function by comparing Game::MSG_ReadBitsCompress and yours (and same for write) and asserting that the outputs are 100% identical. This way even if I don't fully understand it I can at least ensure this will have no undesirable side effects.
Maybe you could include a test function in there that does something like that when iw4x is launched with a `-test-huffman` flag.

Here are some project specific notes:
-  Utils::Huffman::detail should probably be its own file, maybe a header file, so that it doesn't cramp up the Utils::Huffman source file (currently you have to scroll to infinity before you can access any code).
- And then to implement the huffman into the game you would probablby want to create a Component called `Huffman` or something similar, that does the hooking work of plugging it into MSG_ReadbitsCompress/MSG_WriteBittsCompress. If you need directions or offsets I can look into it and provide them for you :) 
- This will also allow the patch to be turned on or off easily from Loader.cpp, should there be any issue.

diamante0018 (approved):

Caball009:
I've set up a couple of (unfinished) unit tests as part of the new huffman component. Would you still like to have a test flag as well, or is it enough to have a unit test?

I'd like to use unit test 2 to compare the output from the old and new huffman code side by side for compression and decompression. I could use some help with the function stubs, though. I'm not sure if we need them, though currently the new huffman code functions expect a size for the output buffer as well, and the game doesn't provide this so that'll need to be sorted out. 

I would also need a way to call the original huffman functions, but I'm not sure if that's possible after installing the hooks.

Rackover:
Looks better!
We have not ran any unit tests on IW4x for a while so I don't know if they even pass, but it's indeed a better idea than a single test flag if they do.
I'm gonna look into your hookings and stuff whenever I can, possibly this weekend. thanks again :)

Caball009:
Sounds good, thank you. Feel free to reach out on Discord if you need anything quickly, though I'll try to keep an eye on this draft.

Caball009:
@Rackover Is there any chance you can have a look at it this weekend? I'd like to write the other unit tests and finish this draft.

Rackover:
Hello, sorry I was not sure this was up for review yet. This week end is a bit complicated for me, I can try looking at it monday night if I don't forget :)  
Don't worry about pinging me if you feel like I'm not answering in reasonabble delays

Caball009:
@Rackover Could you check the new changes, and check if the code works as intended?

https://github.com/iw4x/iw4x-client/pull/111/commits/8efe997cea458cc9488518b0eccd14c7426740b0 removes the huffman exploit fixes, but maybe that's a bit premature, so I'm just making sure that's not overlooked.

Caball009:
I have verified that the new huffman implementation is internally consistent; that is, uncompressed -> compressed -> decompressed yields the same data. The two parts that I haven't verified are the game hooks and the part of unit test 1 that compares the results from the two different huffman implementations.

src/Utils/Huffman.cpp
```diff
@@ -0,0 +1,77 @@
+#include <STDInclude.hpp>
+#include "HuffmanTree.hpp"
+
+namespace Utils::Huffman
+{
+	using namespace Utils::Huffman::Tree;
+
+	int Decompress(const unsigned char* from, unsigned char* to, int fromSize, int toSize)
```
Rackover, line 8: Is there a way that fromSize here (passed from the game) can be negative? If so what will happen?

src/Utils/Huffman.cpp
```diff
@@ -0,0 +1,77 @@
...
+	{
+		assert(fromSize > 0 && toSize > 0);
```
Rackover, line 10: For checks like this it might be better to not rely on assert() but to also have something that works on release (if condition print warning/error and return)
Especially if bad input can/could be provided by a user (malicious or not)

Rackover: This is a pattern that occurs regularly in this PR so I think that for places that could posit a security or stability risk it would be better to not use assert but rather something that works in release too

src/Components/Modules/Security.cpp
```diff
@@ -133,9 +107,6 @@ namespace Components
 	Security::Security()
 	{
 		// Exploit fixes
-		Utils::Hook(0x414D92, Msg_ReadBitsCompressCheckSV, HOOK_CALL).install()->quick(); // SV_ExecuteClientCommands
-		Utils::Hook(0x4A9F56, Msg_ReadBitsCompressCheckCL, HOOK_CALL).install()->quick(); // CL_ParseServerMessage
-
```
Rackover, line 138: I understand that those exploit fixes are no longer necessary with your huffman replacement. 
However due to the component-based nature of IW4x I think it would appreciate if those could remain in case the Huffman component is disabled.
Do you think you could add an `if` here that will restore those security patches if the Huffman component is not present? (for instance you set a static boolean to true in the Huffman constructor, and you check for it here - if it's set to false then those hooks are applied as a security fallback) ?

Rackover (changes requested):
Looks good overall, a few things to address i think and then this needs a live test. If the live test goes through I will be ready to merge this PR

src/Utils/Huffman.cpp
Caball009: I don't think it can happen, but if it did it'd be highly problematic because the buffer size would be enormous as the negative integer becomes a huge number as unsigned integer.

src/Utils/Huffman.cpp
Caball009: I'm not aware that this can be influenced, but I have added checks (instead of assertions) for it. Though if this is a problem, then it has been a problem up until now (in the game's implementation) and will be if the Huffman component is not active.

src/Components/Modules/Security.cpp
Caball009: I added such a boolean to the Huffman component just now. Since the Huffman component is loaded before the Security one, this initialization flag should have been set if the Huffman component is active.

Caball009:
Updated with the requested changes.

[4c0933f](https://github.com/iw4x/iw4x-client/pull/111/commits/4c0933fdfa0c0f7f4bd6174f25f79672e95a8b3a) currently triggers a linker error that I haven't been able to figure out. Is it possible that the Utils::Huffman and Components::Huffman files / namespaces are clashing?

Rackover:
Yes maybe, in that case you can just explicitely specify Components::Huffman when adding it to the loader if you want :)

src/Utils/Huffman.cpp
Caball009: The reason I went with **std::span** is that it has a built-in bounds check (in debug mode), so that any OOB read / write would result in a failed debug assertion at least as opposed to remain hidden. 

If you prefer, I could remove the use of the spans, though, and just rely directly on the byte buffers and signed integer sizes; that way it shouldn't be necessary to check explicitly whether the buffer sizes are > 0.

Caball009:
I already tried that, but it makes no difference.

Caball009:
The linker error can be avoided by renaming either the Huffman util files or the component files btw. Which one should have a different name and what name would that be?

@Rackover Could you check the comments?

Rackover:
So the good thing is that apparently github CI has no problem with your code currently cause it builds & links correctly. I have not tested it myself yet though, and I have not checked the comment yet.
Next week I plan to work a little bit on IW4x and catch up on stuff, so I will be looking at your PR, testing it, and hopefully merging it, and releasing :)

Regarding the discussion of span vs ptr+size_t, I'm not a c++ expert and have never used spans, so I don't know. My opinion is that if a size check is to occur it should occur on debug AND release, so assertions is not protection enough for it 🙂

Caball009:
Even though Github managed to build the code, I cannot get the MSVC linker to do the same thing. So just a heads up, if you ran into the same problem, renaming one set of files (module components or utils) should do the trick.

As for the 'std::span' comments, https://github.com/iw4x/iw4x-client/pull/111/commits/45e8f5489e8a5863c05a290d022427a5c6f33506 is how that could look like (the size checks are now part of the for loop conditions, so they're always checked). As far as I can tell (without having tested the game specific parts), the pr is good to go as-is.

Rackover:
```
^2Huffman decompression out-of-bounds read detected!
^2Huffman decompression out-of-bounds read detected!
0% config string match, 0% config string coverage.
```

Cannot join a single server with your modification! (I tried three different servers)
It is currently not functional, I tested it in debug only but it does not seem to work with existing servers.
Could you please install IW4x and test your code too? This way you can probably find out what's wrong more easily than with my descriptions 🙂

Regarding your linker error: maybe you forgot to run Generate.bat after adding a new file! You need to do it to regenerate the solution. I did it on my end and it worked, and this is what Github CI does too.

Caball009:
You're right, I didn't think to re-generate; that fixes it on my end as well. 

The OOB detection messages are probably expected in debug mode. What's the meaning of the 3rd logged line, and did you check whether the unit tests and hooks were successful? I'm going to check what's required for me test IW4X live btw.

Rackover:
Hello! I'm available on discord if you need help getting IW4x live.  I'm back to work (somewhat) actively on IW4x so I should be able to help you out within reasonable delays if you need anything.

Caball009:
Sounds good. I'll probably contact you at some point on Discord. 

FWIW, I haven't given up on this pr, but it'll take me some additional time before I get around to installing IW4X and do live testing.

Caball009:
It's been a while since my last commit, so has there been any progress?

Rackover:
Hello!  I'm still there, have you managed to set up IW4x to test your pull request on your end? :)
I'm here on discord too if you need help with that. The setup for IW4x is pretty painless so if you can do that and test your PR on live servers to make sure everything works, it's very helpful.
If everything runs as it should I will happily merge this in master and draft a new release! 😃

Caball009:
> Hello! I'm still there, have you managed to set up IW4x to test your pull request on your end? :) I'm here on discord too if you need help with that. The setup for IW4x is pretty painless so if you can do that and test your PR on live servers to make sure everything works, it's very helpful. If everything runs as it should I will happily merge this in master and draft a new release! 😃

I already did that about two months ago after I fixed some issues with the (de)compression. It worked then and it still works now on my end. So it's ready for you to test.

mxve:
I can't comment on code quality and usefulness of this PR, but I did some quick testing.
The game seems to run absolutely fine, I was able to:
- Connect to public servers
- Host a server without and with a mod loaded (botwarfare) & connect to it
- Play private match

Rackover (approved):
Runs well and code looks good, good to go

Rackover:
The game had no issue in Release but testing in debug mode, on a custom map, reveals this issue.
Investigating
image

Caball009:
We already discussed this privately, but I'll put it here as well for posterity. 

The game doesn't have these early break checks for its (de)compression functions, so every time you see these debug prints, it's simply showing how the game would've accessed data beyond the size of the buffer. In other words, the print statements are a reflection of the game's implementation and not of this code, and can be removed if so desired.

If you run unit test 3 and step through it, you should be able to see precisely what's going on here.

src/Components/Modules/Huffman.cpp
```diff
@@ -0,0 +1,184 @@
...
+		Utils::Hook(0x5A85A1, [](bool, const unsigned char* from, unsigned char* to, int fromSize) // CL_Record_f
+			{ return Utils::Huffman::Compress(from, to, fromSize, 0x20000); }, HOOK_CALL).install()->quick();
+		Utils::Hook(0x48FEDD, [](bool, const unsigned char* from, unsigned char* to, int fromSize) // SV_SendMessageToClient
+			{ return Utils::Huffman::Compress(from, to, fromSize, 0x20000); }, HOOK_CALL).install()->quick();
+
+		// Disable the original (de)compression functions
+		// Cannot disable when performing unit tests because they call the original functions
+		if (!Components::Loader::IsPerformingUnitTests())
+		{
+			Utils::Hook(Game::MSG_ReadBitsCompress, [](const unsigned char*, unsigned char*, int) // MSG_ReadBitsCompress
```
diamante0018, line 24: These two hooks weren't properly checked. 
Placing a HOOK_CALL at these addresses will cause the game to crash. This should be a HOOK_JMP. HOOK_CALL will cause the assembly to resume execution of the function after the HOOK was placed because the call instruction pushes to the ESP register the address after the call which is the middle of the function you are hooking. so when your stub returns 0 you will crash.
However, because there is no longer a location in our source code that calls this function and other game addresses were hooked, this condition doesn't seem to happen. However, it should be fixed.

IMO using lambda functions for the body of the hook is also extremely ugly and we never did that for the most part in other parts of our source code.

Rackover: Thank you very much for the code review, although it's a shame it took place the day after merging and not during the ~130 days this pull request was up for review. I will look into it when I can.

Caball009: FWIW, the code for the two hooks disabling the original game functions was only added a week ago, as an afterthought. I'm not very familiar with this hooking library, so my apologies for mixing up HOOK_CALL and HOOK_JMP in these two instances. 

As far as lambda aesthetics go, I can only say that I checked the rest of the codebase and found at least 8 other instances that also used hooking + lambda, and it's not like I added the code like this very recently.

PR #113, opened 2024-05-02
Add BotRemoteAngles and BotAngles

ineedbots:
**What does this PR do?**

Add `BotRemoteAngles` and `BotAngles` builtin methods for bots

Fixes `ads` `BotAction` not throwing (allows bots to shoot akimbo weapons)

Adds `remote` `BotAction`

`BotStop` assigns current weapon

Thanks for the contribution! Please provide a concise description of the problem this request solves.

- [ ] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [ ] Mention any [related issues](https://github.com/XLabsProject/iw4x-client/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [ ] Follow our [coding conventions](https://github.com/XLabsProject/iw4x-client/blob/master/CODESTYLE.md)
- [ ] Minimize the number of commits

### DIFF
--- a/src/Components/Modules/Bots.cpp
+++ b/src/Components/Modules/Bots.cpp
@@ -30,6 +30,8 @@ namespace Components
 		std::uint16_t lastAltWeapon;
 		std::uint8_t meleeDist;
 		float meleeYaw;
+		std::int8_t remoteAngles[2];
+		float angles[3];
 		bool active;
 	};
 
@@ -54,10 +56,11 @@ namespace Components
 		{ "sprint", Game::CMD_BUTTON_SPRINT },
 		{ "leanleft", Game::CMD_BUTTON_LEAN_LEFT },
 		{ "leanright", Game::CMD_BUTTON_LEAN_RIGHT },
-		{ "ads", Game::CMD_BUTTON_ADS },
+		{ "ads", Game::CMD_BUTTON_ADS | Game::CMD_BUTTON_THROW },
 		{ "holdbreath", Game::CMD_BUTTON_BREATH },
 		{ "usereload", Game::CMD_BUTTON_USE_RELOAD },
 		{ "activate", Game::CMD_BUTTON_ACTIVATE },
+		{ "remote", Game::CMD_BUTTON_REMOTE },
 	};
 
 	void Bots::UpdateBotNames()
@@ -214,7 +217,10 @@ namespace Components
 			}
 
 			ZeroMemory(&g_botai[entref.entnum], sizeof(BotMovementInfo));
-			g_botai[entref.entnum].weapon = 1;
+			g_botai[entref.entnum].weapon = ent->client->ps.weapCommon.weapon;
+			g_botai[entref.entnum].angles[0] = ent->client->ps.viewangles[0];
+			g_botai[entref.entnum].angles[1] = ent->client->ps.viewangles[1];
+			g_botai[entref.entnum].angles[2] = ent->client->ps.viewangles[2];
 			g_botai[entref.entnum].active = true;
 		});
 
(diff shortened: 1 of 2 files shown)


src/Components/Modules/Bots.cpp
```diff
@@ -311,6 +317,43 @@ namespace Components
 			g_botai[entref.entnum].meleeDist = static_cast<int8_t>(dist);
 			g_botai[entref.entnum].active = true;
 		});
+
+		GSC::Script::AddMethod("BotRemoteAngles", [](const Game::scr_entref_t entref) // Usage: <bot> BotRemoteAngles(<int>, <int>);
```
Rackover, line 321: It is not immediatly obvious to me what a function called "BotRemoteAngles" does nor what its parameters are, without reading the implementation. It would be nice to document them (the function "sets" the yaw and pitch, even though none of these three word appear in the function name!)

Is there a reason why the "remote angles" is not set at the same time as the "angles" in the other function?
The code does not inform me on why they need to be done separately and the name in the engine ("remote angle") i'm not sure what it means in this context.

src/Components/Modules/Bots.cpp
```diff
@@ -311,6 +317,43 @@ namespace Components
...
+			const auto yaw = std::clamp<int>(Game::Scr_GetInt(1), std::numeric_limits<char>::min(), std::numeric_limits<char>::max());
+
+			g_botai[entref.entnum].remoteAngles[0] = static_cast<int8_t>(pitch);
+			g_botai[entref.entnum].remoteAngles[1] = static_cast<int8_t>(yaw);
+			g_botai[entref.entnum].active = true;
+		});
+
+		GSC::Script::AddMethod("BotAngles", [](const Game::scr_entref_t entref) // Usage: <bot> BotAngles(<float>, <float>, <float>);
+		{
+			const auto* ent = GSC::Script::Scr_GetPlayerEntity(entref);
```
Rackover, line 340: This function can return nullptr !
Please always check for nullptr before accessing the ent's members, lest a bad GSC script could instantly crash the game.

src/Components/Modules/Bots.cpp
```diff
@@ -311,6 +317,43 @@ namespace Components
 			g_botai[entref.entnum].meleeDist = static_cast<int8_t>(dist);
 			g_botai[entref.entnum].active = true;
 		});
+
+		GSC::Script::AddMethod("BotRemoteAngles", [](const Game::scr_entref_t entref) // Usage: <bot> BotRemoteAngles(<int>, <int>);
+		{
+			const auto* ent = GSC::Script::Scr_GetPlayerEntity(entref);
+			if (!Game::SV_IsTestClient(ent->s.number))
```
Rackover, line 324: ditto!

src/Components/Modules/Bots.cpp
```diff
@@ -54,10 +56,11 @@ namespace Components
 		{ "sprint", Game::CMD_BUTTON_SPRINT },
 		{ "leanleft", Game::CMD_BUTTON_LEAN_LEFT },
 		{ "leanright", Game::CMD_BUTTON_LEAN_RIGHT },
-		{ "ads", Game::CMD_BUTTON_ADS },
+		{ "ads", Game::CMD_BUTTON_ADS | Game::CMD_BUTTON_THROW },
```
Rackover, line 59: Would it be okay to map CMD_BUTTON_THROW to a new action named "throw" and then simply set both actions in GSC, therefore managing akimbo in GSC, rather than merging the commands in the client?

Rackover (changes requested):
Some unsafe access of ptr needs to be fixed, otherwise looks okay 👍

src/Components/Modules/Bots.cpp
Rackover: It also returns nullptr if it's not a player!

diamante0018: It's true, it can never be a null pointer because of the long jumps.
Because it was never marked as no return in the definitions at functions.hpp the compiler will complain so those return nullptrs are just there for that reason.
A refactor initiative would be to use c++ exceptions like here  https://git.alterware.dev/alterware/iw6-mod/src/branch/master/src/client/component/gsc/script_extension.cpp#L300

powered by this hook which uses try/catch https://git.alterware.dev/alterware/iw6-mod/src/branch/master/src/client/component/gsc/script_extension.cpp#L56

By using c++ exceptions to handle this in code it would no longer require to leave those nullptrs as the throw statement is already a no return
alternatively leave stuff as it is without any c++ stuff and either comment the behaviour of the function or mark objecterror and all other similar functions as no return

Rackover: I see! :)
Indeed if the function is to never return it should be marked as noreturn and therefore those `return nullptr`  should not be there. Additional short comments (just saying "this long jumps out, no return needed" or something like that) would be much appreciated
The GSC VM is a part of the engine I almost never worked with so I'm just doing due dilligence of checking your implem, sorry if my remarks are a bit naive.
I looked a bit in the engine and the Scr_Error does call Sys_Error which is a noreturn, however there are several conditions to calling Sys_Error (0x2040D06  OR  !0x201A40F)  and they're not documented on my end, I'm not sure what they are.
Do you confirm that Scr_Error will always call Sys_Error and therefore never return?
If so I will trust your word and go through with your change, no problem, but if you're not sure please double check 🙇‍♀️

Rackover (approved):
I'm gonna check up every PR this saturday and draft a new release 💆‍♀️

PR #133, opened 2024-09-02
stringtable: reset not found cache

diamante0018:
This helps modders by overriding novel string tables by making sure that after fs_game is modified and a shutdown is called new tables can be looked up from raw again.

### DIFF
--- a/src/Components/Modules/FileSystem.cpp
+++ b/src/Components/Modules/FileSystem.cpp
@@ -1,4 +1,5 @@
 #include <STDInclude.hpp>
+#include "StringTable.hpp"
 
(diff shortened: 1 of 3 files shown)


Rackover:
How do I test this? 😃

diamante0018:
On second thought, this feature doesn't do everything I wanted it to do.
It allows mods to override CSV files that have not yet been loaded from raw files. However, it would not allow a mod to override CSV files that were already preloaded from the rawfiles that come from IW4x or userraw so this is kinda useless. Don't have time to work on this now. Maybe I'll give it a go another time.

PR #134, opened 2024-09-03
maint(map rotation): add compatibility layer for very old mods

diamante0018:
**What does this PR do?**

I was requested by Xerxes who is now allegedly running this patch on a fork, to add back some code that would allow his plugin to work. This has the advantage of fixing some really old map vote mods as well.

Users can change the map rotation dynamically with `sv_maprotationCurrent` Dvar and have their changes persist even after one map rotation before the Dvar was cleared after just one rotation.

This does not modify or break current setups of any kind and should be a welcomed change.

- [X] Mention any [related issues](https://github.com/XLabsProject/iw4x-client/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [X] Follow our [coding conventions](https://github.com/XLabsProject/iw4x-client/blob/master/CODESTYLE.md)

### DIFF
--- a/src/Components/Modules/MapRotation.cpp
+++ b/src/Components/Modules/MapRotation.cpp
@@ -4,6 +4,8 @@
 #include "MapRotation.hpp"
 #include "Party.hpp"
 
+#define OLD_MAP_ROTATION_BEHAVIOUR
+
 namespace Components
 {
 	Dvar::Var MapRotation::SVRandomMapRotation;
@@ -328,6 +330,10 @@ namespace Components
 		}
 
 		ApplyRotation(rotationCurrent);
+#ifdef OLD_MAP_ROTATION_BEHAVIOUR
+		// We used to allow mods to completely control the rotation through sv_mapRotationCurrent
+		DedicatedRotation = rotationCurrent;
+#endif
 	}
 
 	void MapRotation::SetNextMap(RotationData& rotation)
@@ -371,7 +377,11 @@ namespace Components
 		{
 			Logger::Debug("Applying {}", (*Game::sv_mapRotationCurrent)->name);
 			ApplyMapRotationCurrent(mapRotationCurrent);
+#ifdef OLD_MAP_ROTATION_BEHAVIOUR
+			SetNextMap(DedicatedRotation);
+#else
 			ClearNextMap();
+ #endif
 			return;
 		}
 


Rackover:
I have several questions about this feature 🤔

- Is there any reason why the mods in question cannot be fixed? Like maybe they could use the new system instead (addMap/addGametype) and if you want people to have more control over it you should maybe add a few commands in the new system to allow a similar kind of control?

- Looking at your own comments my friend it looks like that system is supposed to be "deprecated", and has been for a while, so I suppose instead of encouraging the use of a deprecated system you might want to start asking why people are still using the old system instead of the new one. Maybe the new one lacks essential functions the previous one was offering?
image

- If the code from this PR was to ever be released it should probably be controlled by a Dvar instead of a macro? In the worst case, a startup flag? I don't understand why it has to be a macro - i feel like any machine should be perfectly able to build both codes, so using a macro for this sounds out-of-this-world but maybe it's me.

- If people need help updating their mods or map vote mods to the new system we can help them do so on the discord, rather than hacking compatibility ontop of the client **after** treating our users like monkeys for using the old system. What do you think?

Rackover (changes requested):
I have several questions about this feature 🤔

- Is there any reason why the mods in question cannot be fixed? Like maybe they could use the new system instead (addMap/addGametype) and if you want people to have more control over it you should maybe add a few commands in the new system to allow a similar kind of control?

- Looking at your own comments my friend it looks like that system is supposed to be "deprecated", and has been for a while, so I suppose instead of encouraging the use of a deprecated system you might want to start asking why people are still using the old system instead of the new one. Maybe the new one lacks essential functions the previous one was offering?
image

- If the code from this PR was to ever be released it should probably be controlled by a Dvar instead of a macro? In the worst case, a startup flag? I don't understand why it has to be a macro - i feel like any machine should be perfectly able to build both codes, so using a macro for this sounds out-of-this-world but maybe it's me.

- If people need help updating their mods or map vote mods to the new system we can help them do so on the discord, rather than hacking compatibility ontop of the client **after** treating our users like monkeys for using the old system. What do you think?

diamante0018:
I forgot about the code comments, so I guess I'll stick to them since I have authored them.
I have told xerxes that if he wanted the old map rotation system in IW4x then he can open a PR himself but I will no longer pursue this myself.

Rackover:
🙈🙊🙉

PR #137, opened 2024-09-17
Rewrite menu code to track memory usage, prevent crashes and fix leaks

Rackover:
The biggest paradigm change is that menus are now overriden from raw **when the filesystem is reloaded** during a UI_Init or CGameStart. They are no longer loaded on demand, they are always loaded when present.

Normally we should use menu lists (like `code.txt` ) for this, it would be the correct approach.
However this would instantly break decades of mods and servers, so right now I can't consider it. There's a little trick instead to make sure we only load stray menus when in game, and not in the main menu (where most would fail anyway), so mods that modify the main menu will have to have a menulist TXT somewhere to say what should be loaded and what shouldn't.


src/Components/Modules/Menus.cpp
```diff
@@ -507,288 +457,374 @@ namespace Components
...
-				RemoveMenuFromContext(Game::uiContext, list->menus[i]);
+				FreeItem(menu->items[i]);
 			}
 
-			RemoveMenuList(filename);
+			Allocator.free(menu->items);
 		}
 
-		if (Utils::String::EndsWith(filename, ".menu"))
+		FreeExpression(menu->visibleExp);
```
diamante0018, line 704: I believe more data should be freed here as I am trying to compare with that I did with the IW4x-SP menu code

Side by side comparison:

image

src/Components/Modules/Menus.cpp
```diff
@@ -507,288 +457,374 @@ namespace Components
...
+				reallocated->entries = Allocator.allocateArray<Game::expressionEntry>(reallocated->numEntries);
+				std::memcpy(reallocated->entries, statement->entries, sizeof(Game::expressionEntry) * reallocated->numEntries);
+			}
 
-		for (auto i = MenuList.begin(); i != MenuList.end(); ++i)
-		{
-			FreeMenu(i->second);
+			if (andFree)
+			{
+				Game::free_expression(statement);
```
diamante0018, line 666: I believe here a recursive call to free -> supportingData (If necessary) is missing. This would be necessary if IW4x was to support parsing UI functions. It may not be necessary at the moment but could be documented nonetheless.

image

diamante0018 (commented):
I tried to compare it against the code that I wrote on IW4x-SP.

Until I did all the memory *DE*Allocation code  I could note a memory increase after every single map rotate when I had custom menus loaded from RAW which is what you seem to be reporting in the discord channels. I believe this is still tied to a menu memory leak.

https://git.alterware.dev/alterware/iw4x-sp/src/branch/master/src/client/component/ui.cpp

src/Components/Modules/Console.cpp
```diff
@@ -58,10 +58,10 @@ namespace Components
 
 	bool Console::isCommand;
 
-	const char** Console::GetAutoCompleteFileList(const char* path, const char* extension, Game::FsListBehavior_e behavior, int* numfiles, int allocTrackType)
+	const char** Console::GetAutoCompleteFileList(const char* path, const char* extension, Game::FsListBehavior_e behavior, int* numfiles, [[maybe_unused]] int allocTrackType)
 	{
 		if (path == reinterpret_cast<char*>(0xBAADF00D) || path == reinterpret_cast<char*>(0xCDCDCDCD) || ::Utils::Memory::IsBadReadPtr(path)) return nullptr;
-		return Game::FS_ListFiles(path, extension, behavior, numfiles, allocTrackType);
+		return Game::FS_ListFiles(path, extension, behavior, numfiles);
```
diamante0018, line 64: The canon function does take a 5th parameter, so removing it from a function definition isn't ideal. It's unused but I would still pass it to keep the stack the same the game expects it to be.

src/Components/Modules/QuickPatch.cpp
```diff
@@ -298,8 +298,32 @@ namespace Components
...
+		}
+		else
+		{
+			return nullptr;
+		}
+	}
+
 	QuickPatch::QuickPatch()
 	{
+		Utils::Hook(0x4EC1D0, SL_ConvertToStringHk, HOOK_JUMP).install()->quick();
```
diamante0018, line 324: just gonna assume this is subject to change and not going to be merge as is.

diamante0018 (commented):
You seem to unnecessarly Include the components namespace when invoking the logger. (a copy paste from the zone/asset code) You are already in the components namespace most of the time so it's not needed

diamante0018:
btw this is how guy debugs SL and Memtree now: https://gist.github.com/ineedbots/ccd66cf076e67b6033bea8fb8d965bc8

src/Components/Modules/QuickPatch.cpp
Rackover: Yea woops

src/Components/Modules/Console.cpp
Rackover: Am gonna check again but I was pretty sure it was not in the engine, if it's unused i'd rather not add it to the API of IW4x (otherwise everytime you call the function you're like, "what's this again?")

src/Components/Modules/Menus.cpp
Rackover: addressed that for everything but the key handler, i'll open anotehr issue for the leaks when this is merged

src/Components/Modules/Console.cpp
Rackover: I gave it a thought and decided it was okay - the parameter is unused so it must never be supplied. Having it in the API creates noise for the developer and has no otherwise benefit than accuracy, something IW4x is not really interested in, so I'll leave that parameter out. I might remove it from other places too for consistency (FS_List is not the only one having it around)

Rackover:
I've tested it, so I'm merging and releasing it right now because this addresses a very frequent crash. I will open another pr for next changes if needs be

src/Components/Modules/Console.cpp
diamante0018: To say it "must never be supplied" is like, not accurate. It's the fact that it is unused that makes it okay to not supply an extra argument that is not used in the function. But the game itself still supplies it so it's completely legal to supply it. So yeah to say it must never be supplied is technically wrong.

image

It's a choice not to supply it and to remove it from the game definition.

Rackover: No no, it must never be supplied in the IW4x code, this is what I mean.
As a developer of IW4x, there is no case where you must supply it to use the function, so, it must never be supplied. Leaving open the option for the developer to supply it, is a mistake. It's like adding a button on the control panel that does nothing - it will only lead to confusion. So it must never be supplied. That's my stance :)

I've been looking at the same assembly as you and yes I saw the game pushes `10` before calling the function, and not just this one, despite that it's never used, so I know that "for accuracy" you could want to open that parameter. But for reasons I hope I made clear prior, I think it does not belong in the game API of IW4x, because it must never be supplied.

PR #142, opened 2024-09-22
Fix standard attributes in middle of decl-specifiers

wroyca:
Attributes can only appear first in a member declaration. MSVC is lenient here, but other compilers will generate an error accordingly.

https://eel.is/c++draft/class.mem#general


Rackover:
I'm not even sure why we have those. Somebody said it's good practice but then didn't update the whole project to use them so now they're in this tiny place and nowhere else

wroyca:
Will make a follow-up PR to remove these. :)

PR #148, opened 2024-10-01
Clip cursor to the game window.

wroyca:
This PR addresses a bug where the cursor could move outside the game window during a flick (a very quick mouse movement to reposition the crosshair). It should also prevent the mouse cursor from "jumping" to second monitor.

### DIFF
--- a/src/Components/Modules/RawMouse.cpp
+++ b/src/Components/Modules/RawMouse.cpp
@@ -82,11 +82,6 @@ namespace Components
 
 		if (GetForegroundWindow() == Window::GetWindow())
 		{
-			if (r_fullscreen.get<bool>())
-			{
-				IN_ClampMouseMove();
-			}
-
 			static auto oldX = 0, oldY = 0;
 
 			auto dx = MouseRawX - oldX;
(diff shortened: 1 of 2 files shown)


diamante0018 (approved):

SnowyWhite:
I've tested it on my dual-monitor setup and compared it to the current release version ([r4625](https://github.com/iw4x/iw4x-client/releases/tag/r4625)).
It was possible for me to reproduce the bug in the release version by 
- disabling fullscreen mode
- disabling window border
- setting the video resolution to 1920x1080 (my display is 3440x1440).

If I moved the mouse quickly enough, it would occasionally jump to my second monitor. 
With the same settings and your fix applied, I can confirm that this no longer happens! 
It also didn't happen when I changed any of the video settings and I didn't see any side effects.

Rackover:
Tested it a bit and it looks like it's working with no side effects
I don't think `IN_RecenterMouse` is supposed to return a BOOL (this looks accidental as it is never used) but it is not very important 🙂

Rackover (approved):

PR #150, opened 2024-10-07
fix: Add git tags if missing

SnowyWhite:
**What does this PR do?**

This PR adds a small fix to the `generate.bat` script.

If you fork this repository, then GitHub, by default, only copies the `develop` branch: 
image

The problem here is that the [premake5.lua](https://github.com/iw4x/iw4x-client/blob/34594512a10961e9b4f9b5a3ff72ba7e989c73ef/premake5.lua#L117) script requires git tags in order to generate the build information file.
Because of this, compiling a forked clone with the default settings will fail.

The easiest fix to this problem is to simply add the original repository as an upstream remote and then fetching / pushing the git tags to the fork. That's exactly what this PR does:
It first checks if there is an upstream remote, and if not, adds it and copies the git tags.

It would probably be enough to just copy the most recent tag. 
However, for the sake of completeness, it will copy them all once when it generates the solution.

Another solution would be to edit the lua script, but I have no experience with Lua and, I believe, it would require the creation of a new tag, which could potentially have some caveats.

### DIFF
--- a/generate.bat
+++ b/generate.bat
@@ -1,4 +1,16 @@
 @echo off
 echo Updating submodules...
 call git submodule update --init --recursive
+
+:: Check if the repository has an upstream remote (indicating a fork)
+git remote get-url upstream >nul 2>&1
+IF ERRORLEVEL 1 (
+    echo No upstream remote found. This might be a fork.
+    echo Adding upstream...
+    git remote add upstream https://github.com/iw4x/iw4x-client.git
+    echo Adding git tags...
+    git fetch upstream --tags
+    git push origin --tags
+)
+
 call tools\premake5 %* vs2022


generate.bat
```diff
@@ -1,4 +1,16 @@
...
+
+:: Check if the repository has an upstream remote (indicating a fork)
+git remote get-url upstream >nul 2>&1
+IF ERRORLEVEL 1 (
+    echo No upstream remote found. This might be a fork.
+    echo Adding upstream...
+    git remote add upstream https://github.com/iw4x/iw4x-client.git
+    echo Adding git tags...
+    git fetch upstream --tags
+    git push origin --tags
```
mxve, line 13: I don't know if the tags should get pushed to the remote, pushing should always be a manual thing IMO

Laupetin:
Wouldn't it be also an option to make generating build information in premake work without tags? If there are none it could use a hardcoded value like "fork" or "dev" or whatever since i guess if you dont copy tags when forking it might be okay to not have the latest tag information.

generate.bat
SnowyWhite: My reasoning was that you probably want a fork to be an exact copy of a repository anyway (I don't like GitHub's default option), so there's no harm in pushing the tags into the fork. But if just fetching the tags is enough to successfully compile the project, then of course this can be changed.

diamante0018:
I think the premake5 script needs to be edited indeed

mxve:
I think the tag-based versions aren't used anywhere anymore, so maybe we should get rid of them altogether.

https://github.com/iw4x/iw4x-client/pull/145

Rackover:
So what do we think? 
We pull the tags as part of generate or we stop using the tags in premake entirely?

If they're not useful I'm up for not using them in premake, or making them facultative so that nobody's branch breaks when they remove all tags (you wouldn't expect _that_ to break a build)

Let me know so we can merge that in before next release

wroyca:
> So what do we think? We pull the tags as part of generate or we stop using the tags in premake entirely?
> 
> If they're not useful I'm up for not using them in premake, or making them facultative so that nobody's branch breaks when they remove all tags (you wouldn't expect _that_ to break a build)
> 
> Let me know so we can merge that in before next release

This has been a persistent issue every time I need to check out or test outside upstream. Let's remove it altogether—the entire pushd machinery needs to go

wroyca:
Closing with https://github.com/iw4x/iw4x-client/pull/205

PR #152, opened 2024-10-10
update .github

### LINKED ISSUE #115: Update invalid link to code style page in pull request template
The current template for pr's contains a link to https://github.com/XLabsProject/iw4x-client/blob/master/CODESTYLE.md instead of https://github.com/iw4x/iw4x-client/blob/master/CODESTYLE.md

mxve:
- Resolves #115 
- Removed CODEOWNERS
  - From what I understand this is used to automatically assign reviewers. Using @iw4x/Developers does not work as its not public.
- Formatting fix on build.yml
- Fixed typo in README that was driving me insane

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
           - Release
     steps:
       - name: Check out files
-        uses: actions/checkout@v3.5.3
+        uses: actions/checkout@v4
         with:
           submodules: true
           fetch-depth: 0
@@ -32,7 +32,7 @@ jobs:
           lfs: false
 
       - name: Add msbuild to PATH
-        uses: microsoft/setup-msbuild@v1.3.1
+        uses: microsoft/setup-msbuild@v2
 
       - name: Generate project files
         run: tools/premake5 vs2022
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
diamante0018, line 1: Please update everything here to use `@ v4` of github actions. using nodejs16 will throw warnings and is deprecated

diamante0018 (changes requested):

SnowyWhite:
The [SECURITY.md](https://github.com/iw4x/iw4x-client/blob/develop/SECURITY.md) file has the mail address of momo, I think this PR would be a great opportunity to fix this too!

diamante0018:
> The [SECURITY.md](https://github.com/iw4x/iw4x-client/blob/develop/SECURITY.md) file has the mail address of momo, I think this PR would be a great opportunity to fix this too!

Should be updated by a maintainer/project owner and not via a PR although I wouldn't know.. Perhaps open an issue about this for Louvenarde to see

diamante0018 (commented):
updated build file looks good to me. this should get rid of the warnings from the actions tab

Rackover (approved):

PR #153, opened 2024-10-11
Discard non-existent dvar

enclave:
**What does this PR do?**

This commit removes all references to it and uses `aim_lockon_strength` as the scale/strength factor when calculating the lock-on pitch turn rate. This aligns with the behavior in the Xbox 360 version of the game.

I am unsure as I do not have a controller to test.

The dvar `aim_lockon_pitch_strength` exists in T5 (I believe from recollection). Whether there should be seperate dvars for lock-on pitch and yaw strength is another question, and if so, the dvars should then be registered properly.

### DIFF
--- a/src/Components/Modules/Dvar.cpp
+++ b/src/Components/Modules/Dvar.cpp
@@ -259,6 +259,11 @@ namespace Components
 		return Game::Dvar_RegisterFloat(dvarName, value, min, 10000.0f, flags, description);
 	}
 
+	const Game::dvar_t* Dvar::Dvar_RegisterAimLockonStrength(const char* dvarName, float value, float min, float max, std::uint16_t flags, const char* /*description*/)
+	{
+		return Game::Dvar_RegisterFloat(dvarName, value, min, max, flags, "The amount of aim assistance given by the target lock on (yaw)");
+	}
+
 	void Dvar::SetFromStringByNameSafeExternal(const char* dvarName, const char* string)
 	{
 		static std::array exceptions =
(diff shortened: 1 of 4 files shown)


diamante0018:
Good catch, yeah IW4x controller support was pasted from T5 and not IW4 Xbox so mistakes like this happened.

Rackover:
Is there a reason why we would want to discard this variable?

It allows managing the pitch correction separately from the yaw correction. Sure, it might not exist in IW4 Xbox. But IW4x contains a lot of other stuff that does not exist on IW4 Xbox... ! I'm not sure how this is relevant.

Could you describe in details what problem this PR intends to solve? 🤔

enclave:
This PR will bring parity to how the lock-on turn rates are calculated on the Xbox 360 version of the game using a single dvar (`aim_lockon_strength`) as the strength/scale factor.

While you're correct that this does not need to be followed just because it's done that way on the Xbox 360 version, if a separate dvar such as `aim_lockon_pitch_strength` is desired, then the dvar should be registered properly. As the dvar does not actually exist since it's not registered by the game, this [line](https://github.com/iw4x/iw4x-client/blob/develop/src/Components/Modules/Gamepad.cpp#L1885) of code will consequently register an empty string dvar,  due to the implementation of [`Dvar::Var`](https://github.com/iw4x/iw4x-client/blob/develop/src/Components/Modules/Dvar.cpp#L10:L18). Therefore, in the provided screenshot you are able to see how this dvar is missing an assigned value and more.

image

Rackover:
I understand better now! 🙂  Thansk for the explanations
In that case I would rather have this PR add a registration for this Dvar, so that it is valid to the game and displayed with the correct tyype / boundaries / restriction,  instead of removing the Dvar altogether, if that's okay

Rackover:
Tested and works

Rackover (approved):

PR #154, opened 2024-10-13
[PlayerMovement] Tackle some issues with Dvar flags

### LINKED ISSUE #147: Uncheat bg_playercollision
**What problem will this solve?**
I have an issue with my server where I can turn off bg_playercollision from the server side but not from the client side (at is it now a protected dvar), so when passing through other players there is a lag issue.

**What might be a solution?**
Put back **bg_playercollision** (and maybe bg_playerEjection) to Game::DVAR_CODINFO and not DVAR_CHEAT anymore.

**What other alternatives have you already considered?**
Made a custom build (with the source code) for my server players. (which is not perfect because ill have to make a new one on every new update/release).

No.

diamante0018:
Closes #147 Dvars that are server side do not need to be cheat protected because the server will correct any out of sync client

### DIFF
--- a/src/Components/Modules/PlayerMovement.cpp
+++ b/src/Components/Modules/PlayerMovement.cpp
@@ -321,26 +321,26 @@ namespace Components
 			"The speed at which noclip camera moves");
 
 		BGDisableLandingSlowdown = Game::Dvar_RegisterBool("bg_disableLandingSlowdown",
-			false, Game::DVAR_CHEAT, "Toggle landing slowdown");
+			false, Game::DVAR_CODINFO, "Toggle landing slowdown");
 
 		BGBunnyHopAuto = Game::Dvar_RegisterBool("bg_bunnyHopAuto",
-			false, Game::DVAR_CHEAT, "Constantly jump when holding space");
+			false, Game::DVAR_CODINFO, "Constantly jump when holding space");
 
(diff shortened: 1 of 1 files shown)


diamante0018:
@Rackover kindly requesting a review

Rackover:
This PR does not need further justification, I simply asked for clarification about Dvar flags, which this PR changes. 
Imo this change makes sense, I just want to make sure we're not uncheating something we shouldn't - and _also_, considering whether other Dvars should also be uncheated along with those.

diamante0018:
> This PR does not need further justification, I simply asked for clarification about Dvar flags, which this PR changes. Imo this change makes sense, I just want to make sure to understand what each flag is for as to make sure we're not uncheating something we shouldn't - and _also_, considering whether other Dvars should also be uncheated along with those.

No problem, I did try to put some very basics comments in the structs header files alongside the dvar flag definition but now I saw that there is a wiki so perhaps all dvars can be expanded on there with lengthier explanations because not all behaviour is documented. Some dvar flags seem unused too like what was labelled as DVAR_INTERNAL

Yes part of issue 61 was original a request to uncheat a specific dvar that is server-side (BG) however the list is probably longer and it could be a matter of uncheating flags upon request when a need is justified or it is brought up in a feature request.

Rackover (approved):

Issue #61, opened 2023-10-26
Uncheat player_sprintunlimited


diamante0018:
**What problem will this solve?**
Marathon perk for everyone without using devmap

**What might be a solution?**
Uncheat player_sprintunlimited

**What other alternatives have you already considered?**
There are none because using devmap allows players to use other unwanted cheats

**Anything else we should know?**
No

Rackover:
After discussions on the discord we advised that the better alternative would be to unprotect SV_CHEATS and allow server owners to change it at runtime just like in source engine, but this would need a bit of testing and/or code as the value might be cached
TODO

diamante0018:
A lot of dvars where unnecessarily cheat protected in playermovement.cpp

In other words, dvars that were completely server-sided were cheat protected.

https://github.com/iw4x/iw4x-client/blob/develop/src/Components/Modules/PlayerMovement.cpp#L306-L342

Only recoil dvars needs to be cheat protected. the code change should be reverted and the other dvars should return to codinfo flag

diamante0018:
server owners can't disable player collision anymore without running devmap which is incredibly hard to work with because enabling sv-cheats allows players to do a bunch of other stuff like using god commands. Server owners need to be able to freely edit the highlighted dvars without enabling sv_cheats

diamante0018:
bg_recoilmultiplayer's dvar name should be change. it is an incedibly misleading dvar name, BG = both games, it runs on the client only, name  should be changed to cg_

diamante0018:
> After discussions on the discord we advised that the better alternative would be to unprotect SV_CHEATS and allow server owners to change it at runtime just like in source engine, but this would need a bit of testing and/or code as the value might be cached TODO

I agree with this, I have now realized this is implemented on iw6-mod and all of Plutonium clients via jump hooks in dvar_variant function. I will try to link that here

Rackover:
![both-is](https://github.com/user-attachments/assets/009d46bd-b1f6-45e1-a826-1343a3dd9bac)

diamante0018:
Requirements.
```
In other games, as far as I know, the host is held to the same "cheat" policy as the clients
If a client or a host wants to edit a cheat dvar, they need sv_cheats to be set to 1, which is in control of the host
The reason sv_cheats is in control of the host is simply because, well, it's a server variable, so the host controls it
The problem so far was that CoD locks-in the sv_cheats variable the moment a server is started, which is annoying for the host
[...]

PR #155, opened 2024-10-13
fix dvar name in player movement

diamante0018:
Addresses parts of #61 (do not close yet)

Breaking: change dvar name (necessary to avoid bad confusion as it did in the past: bg(both games) name will lead someone who is familiar with quake engine to think that dvar is checked by server side when it is NOT and must cheat protected because it is CG category (client game)

Yes: above there is an explanation

no

### DIFF
--- a/src/Components/Modules/PlayerMovement.cpp
+++ b/src/Components/Modules/PlayerMovement.cpp
@@ -12,7 +12,6 @@ namespace Components
 	const Game::dvar_t* PlayerMovement::BGPlayerEjection;
 	const Game::dvar_t* PlayerMovement::BGPlayerCollision;
 	const Game::dvar_t* PlayerMovement::BGClimbAnything;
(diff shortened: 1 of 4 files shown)


Rackover:
i'm thinking maybe this should have been Weapon.cpp too . . .   i  didn't really put it in the right place at all

diamante0018:
I see, would you like me to move the dvar via this pull request to the other cpp file?

Rackover:
Yes if that's OK i think that would be a good idea while we're here

diamante0018:
Waiting on CI... All is done

diamante0018:
fixed my PR should be good now

Rackover:
will be testing it shortly and then i'll merge it

diamante0018:
Did testing go well?

Rackover:
i don't know, did it?

Rackover:
Works as intended

Rackover (approved):

PR #156, opened 2024-10-17
[Mods]: Clear fs_game when leaving servers

### LINKED ISSUE #139: Reset fs_game when leaving a game
This prevents creating a mod that fucks up the main menu in a way that the users dont know how to quit the game or remove the mod!
Modders have been using this massively, and having everybody relaunch the game after leaving a server is a bug.
It's an exploit and should be considered as such, as it is always to the detriment of the user :(

diamante0018:
**What does this PR do?**

Closes #139 

Unloads the mod, if loaded, when leaving a server
Fixes some bugs I spotted(unterminated Cbuf strings) in the mod list code

As explained in the code #70 does not allow me to unload a mod when leaving a private match

### DIFF
--- a/src/Components/Modules/Download.cpp
+++ b/src/Components/Modules/Download.cpp
@@ -5,6 +5,7 @@
 #include "Download.hpp"
 #include "Events.hpp"
 #include "MapRotation.hpp"
+#include "ModList.hpp"
 #include "Node.hpp"
 #include "Party.hpp"
 #include "ServerInfo.hpp"
@@ -348,7 +349,7 @@ namespace Components
 
 				Command::Execute("closemenu mod_download_popmenu", false);
 
-				if (Dvar::Var("cl_modVidRestart").get<bool>())
+				if (ModList::cl_modVidRestart.get<bool>())
 				{
 					Command::Execute("vid_restart", false);
 				}
--- a/src/Components/Modules/ModList.cpp
+++ b/src/Components/Modules/ModList.cpp
@@ -1,4 +1,6 @@
 #include <STDInclude.hpp>
+
+#include "Events.hpp"
 #include "ModList.hpp"
 #include "UIFeeder.hpp"
 
@@ -65,11 +67,11 @@ namespace Components
 
 		if (Dvar::Var("cl_modVidRestart").get<bool>())
 		{
-			Game::Cmd_ExecuteSingleCommand(0, 0, "vid_restart");
+			Command::Execute("vid_restart", false);
 		}
 		else
 		{
-			Game::Cmd_ExecuteSingleCommand(0, 0, "closemenu mods_menu");
+			Command::Execute("closemenu mods_menu", false);
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


diamante0018:
Hi @Rackover, main feature test showcase on YouTube (Link sent in DMs)

Private Match tested off-camera but works as intended and is explained above.

Rackover:
Is there a reason why ou don't call ClearMods as you originally intended? 🤔 
It looks like it would do the trick for your case. Doesn't it?

diamante0018:
Yes. See comment in code. Besides clear mods is a lambda function so anonymous functions can't be invoked directly.

diamante0018:
Clear mods is used by the menu code, it's a UIScript. So it's only displayed when fs_game is set.

https://private-user-images.githubusercontent.com/37080671/377546103-0b4943f3-4c3e-48fa-b132-452abb3f595a.png

So the user has to manually trigger this when going to the menu if a mod was loaded for this to display.

Now this PR is about unloading a mod once leaving a server so a few things need to be checked out

- fs_game is set
- we are not leaving a private game
- unload the mod without checking `cl_modVidRestart`

all reasons as to why code from clearmod cannot be reused.

i invite you to check out the clearmod as it's displayed in the iw4x menus as it should make sense why there checking for private match wouldn't make sense (we are on the main menu)

Desktop Screenshot 2024 10 18 - 10 32 28 84

Rackover:
Yes, I know where that button is and what function is tied to.
This simply does not answer my questions:

- Why do you need to "unload the mod without checking cl_modVidRestart" ... ? If cl_modVidRestart is set to a certain value, that value should be respected. When you join a server with cl_modVidRestart set to false, you don't get a vid restart (see Download.cpp) so why should you get one when leaving...? can you explain that to me?
- I'm sure you're aware that in C++, you can create new functions when you need to. If the UIScript ClearMods, and the leave server event, have functionality _in common_, such functionality can be put in a new function (UnloadMod, or ClearMod, or whatever you desire) and then that function can be called from both the UIScript lambda and from the leave server event. 

Reduce, reuse, recycle - less duplicates of codes that does the something (why would "new code" be preferable to "old code"? does code rot?), less stuff to maintain - only one correct way to unload a mod, one function to call, and you can add the local server check specificity to your event if you want **before** calling that function. You know all this! Why pretend you don't, and have me lay it all out? ☹️

Rackover:
Additionally - have you checked that `->modified = true` needs to be set after calling Dvar_SetString?
Im not really up to date on Dvars so I might be wrong on this. But i feel like this modified flag is set by Dvar_SetString at `0x6478B7`. Tell me if this is mistake on my end

diamante0018:
I see 0x6478B7, I guess it is indeed unnecessary and a bad copy and paste on my end.

Well, if that's what you desire, I will now modify the PR to do what you are suggesting.

diamante0018:
I have completed testing once again. All is good on my end.

On a side note, it's impossible to find a server that does not have a gazillion menu errors ahah. I'm will be so glad once this PR is merged because who knows how broken their menus are outside of normal gameplay

Desktop Screenshot 2024 10 18 - 13 57 07 18

Rackover:
unrelatedly - there's a crash currently when you leave a game and then try to open the server browser. It's very old and i'm gonna open an issue for it because it definitely hinders usability (currently the game crash if i leave a game then i click "join game")

Rackover (approved):

Rackover:
Tested and it seems to run gracefully, through map rotation aswell, merging

PR #158, opened 2024-10-24
Fix editorconfig misconfiguration

wroyca:
This PR fix the current `.editorconfig` by designating it as the "top-most" configuration. The previous *(faulty)* behavior caused editors to search for `.editorconfig` files upwards in the directory tree until encountering one with `root=true`, which would then override settings from our (IW4x) `.editorconfig`.

I've also introduced a rule to preserve trailing whitespace in Markdown, as Markdown requires two or more spaces at the end of a line to create a `<br />` tag for a line break (refer to my comment in the `.editorconfig` file from this PR).

Note that code-style settings (e.g., `cpp_indent_namespace_contents`) are removed. They are extensions and not officially supported by `.editorconfig` itself. If we want to enforce a proper style convention, we should delegate that to the appropriate tooling (in this case, [clang-format](https://clang.llvm.org/docs/ClangFormatStyleOptions.html))

Note (second one) that `indent_size` was also removed. That is because IW4x uses tabs for indentation, meaning that `indent_size` can be set to any value in the editor based on personal preference, without affecting the actual code (beyond local visual appearance).

### DIFF
--- a/src/Components/Modules/Gamepad.cpp
+++ b/src/Components/Modules/Gamepad.cpp
@@ -174,6 +174,7 @@ namespace Components
 
 	GamepadControls::Controller Gamepad::gamePads[Game::MAX_GPAD_COUNT]{};
 	std::mutex Gamepad::gamePadStateMutexes[Game::MAX_GPAD_COUNT]{};
+	bool Gamepad::gamePadDataReady[Game::MAX_GPAD_COUNT]{};
 	Gamepad::GamePadGlobals Gamepad::gamePadGlobals[Game::MAX_GPAD_COUNT]{ {} };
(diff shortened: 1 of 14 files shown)


diamante0018 (approved):

Rackover:
was thinking on whether there was a wa to also commit all the current IW4x source files, reformatted using _these_ settings, in a separate commit on the same PR - so we see what it looks like and get an updated source format immediatly

wroyca:
> was thinking on whether there was a wa to also commit all the current IW4x source files, reformatted using _these_ settings, in a separate commit on the same PR - so we see what it looks like and get an updated source format immediatly

Done :)

PR #160, opened 2024-10-27
fix: make .str parsing with ZB useful

diamante0018:
**What does this PR do?**

Makes CoD4 str parsing exclusive to a flag. Undoing this default behaviour will make ZB modder friendly and can allow Louvenarde to undo the JSON localized string parsing from the IW4x rawfiles

No longer adds prefix to localized strings derived from .str file name (CoD4 mod tools behaviour)

No

### DIFF
--- a/src/Components/Modules/AssetInterfaces/ILocalizeEntry.cpp
+++ b/src/Components/Modules/AssetInterfaces/ILocalizeEntry.cpp
@@ -54,32 +54,34 @@ namespace Assets
 	}
 
 	void ILocalizeEntry::ParseLocalizedStringsFile(Components::ZoneBuilder::Zone* builder, const std::string& name, const std::string& filename)
-	{
-		std::vector<Game::LocalizeEntry*> assets;
-		const auto _0 = gsl::finally([]
+	{		
+		if (Components::Flags::HasFlag("-original-str-parsing"))
 		{
-			Components::Localization::ParseOutput(nullptr);
-			Components::Localization::PrefixOverride = {};
-		});
-
-		Components::Localization::PrefixOverride = Utils::String::ToUpper(name) + "_";
-		Components::Localization::ParseOutput([&assets](Game::LocalizeEntry* asset)
-		{
-			assets.push_back(asset);
-		});
+			Components::Localization::PrefixOverride = Utils::String::ToUpper(name) + "_";
+		}
 
 		const auto* psLoadedFile = Game::SE_Load(filename.data(), false);
 		if (psLoadedFile)
 		{
 			Game::Com_PrintError(Game::CON_CHANNEL_SYSTEM, "^1Localization ERROR: %s\n", psLoadedFile);
+			Components::Localization::PrefixOverride = {};
 			return;
 		}
+
+		std::vector<Game::LocalizeEntry*> assets;
+		Components::Localization::ParseOutput([&assets](Game::LocalizeEntry* asset)
+		{
+			assets.push_back(asset);
+		});
 
 		auto type = Game::DB_GetXAssetNameType("localize");
 		for (const auto& entry : assets)
 		{
 			builder->addRawAsset(type, entry);
 		}
+
+		Components::Localization::ParseOutput(nullptr);
+		Components::Localization::PrefixOverride = {};
 	}
 
 	void ILocalizeEntry::ParseLocalizedStringsJSON(Components::ZoneBuilder::Zone* builder, Components::FileSystem::File& file)
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ To use the `iw4x.dll`, you must have a valid Modern Warfare 2 installation with 
 | `-nosteam`              | Disable friends feature and do not update Steam about the game's current status just like an invisible mode. |
 | `-unprotect-dvars`      | Allow the server to modify saved/archive dvars. |
 | `-zonebuilder`          | Start the interactive zonebuilder tool console instead of starting the game. |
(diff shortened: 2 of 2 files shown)


src/Components/Modules/AssetInterfaces/ILocalizeEntry.cpp
```diff
@@ -62,7 +62,12 @@ namespace Assets
 			Components::Localization::PrefixOverride = {};
 		});
 
-		Components::Localization::PrefixOverride = Utils::String::ToUpper(name) + "_";
+		
+		if (Components::Flags::HasFlag("-original-str-parsing"))
+		{
+			Components::Localization::PrefixOverride = Utils::String::ToUpper(name) + "_";
```
Rackover, line 68: I was confused for a moment because of the GSL finally block above - i thought the PrefixOverride was never reset but there's actually an exit block that executes on destruction right above.


however the GSL Finally exit block looks like it's not useful - when the SE_Load call under this line fails, the whole function returns without doing anything to zonebuilder, and so the whole Localization::ParseOutput and PrefixOverride assignation is useless when that happens.

If that's okay I would like you to move things around a little bit to do the `SE_Load` **first**  - if it fails, there's an early return that does not need any sort of `finally` block to execute. Else, you can just put the existing code underneath it to execute sequentially - no need for gsl::finally, and since the code will be written in the order it's executed it will be easier to read too.

```
function()
{
  if (!SE_Parse)
  {
     return;
  }
  
  Localization::Parse(...);
  Prefix = ... ;
  
  for (...)
  {
     builder->addRawAsset(type, entry);
  }
  
  Prefix = {};
  Localization::Parse(nullptr);
end
```

Rackover (changes requested):

diamante0018:
I should have addressed your review @Rackover

diamante0018:
This PR can be tested using the other PR I opened on the rawfiles repository btw as it's the first PR that reintroduces .str files to the rawfiles repo as they were previously removed.

Rackover:
I wanted to put the SE_Load fail earlier so that it would return early in the function without modifying state, as per my review comment.

However despite that "Maintainers are allowed to edit this pull request", for some reason, I can't push to that remote ☹️ 

```
Enumerating objects: 13, done.
Counting objects: 100% (13/13), done.
Delta compression using up to 16 threads
Compressing objects: 100% (7/7), done.
Writing objects: 100% (7/7), 599 bytes | 199.00 KiB/s, done.
Total 7 (delta 6), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (6/6), completed with 6 local objects.
To github.com:diamante0018/iw4x-client-strings.git
 ! [remote rejected]   diamante/zb_fix -> fix/zb_str (permission denied)
error: failed to push some refs to 'github.com:diamante0018/iw4x-client-strings.git'
```

So here you go, the hard way 😂 
```diff
---
 .../Modules/AssetInterfaces/ILocalizeEntry.cpp     | 14 +++++++-------
 1 file changed, 7 insertions(+), 7 deletions(-)

diff --git a/src/Components/Modules/AssetInterfaces/ILocalizeEntry.cpp b/src/Components/Modules/AssetInterfaces/ILocalizeEntry.cpp
index 33ae9a2f..81ae988f 100644
--- a/src/Components/Modules/AssetInterfaces/ILocalizeEntry.cpp
+++ b/src/Components/Modules/AssetInterfaces/ILocalizeEntry.cpp
@@ -55,6 +55,13 @@ namespace Assets
 
 	void ILocalizeEntry::ParseLocalizedStringsFile(Components::ZoneBuilder::Zone* builder, const std::string& name, const std::string& filename)
 	{
+		const auto* psLoadedFile = Game::SE_Load(filename.data(), false);
+		if (psLoadedFile)
+		{
+			Game::Com_PrintError(Game::CON_CHANNEL_SYSTEM, "^1Localization ERROR: %s\n", psLoadedFile);
+			return;
+		}
+
 		std::vector<Game::LocalizeEntry*> assets;
 		
 		if (Components::Flags::HasFlag("-original-str-parsing"))
@@ -67,13 +74,6 @@ namespace Assets
 			assets.push_back(asset);
 		});
 
-		const auto* psLoadedFile = Game::SE_Load(filename.data(), false);
-		if (psLoadedFile)
-		{
-			Game::Com_PrintError(Game::CON_CHANNEL_SYSTEM, "^1Localization ERROR: %s\n", psLoadedFile);
-			return;
-		}
-
 		auto type = Game::DB_GetXAssetNameType("localize");
 		for (const auto& entry : assets)
 		{

```
[patch.zip](https://github.com/user-attachments/files/17990919/patch.zip)

Rackover:
I'm surprised, does that means the asset is supposed to be pushed even in case of load error?

diamante0018:
Im not too sure I get that. There is a return statement when an error occurs. Therefore the assets won't get pushed, right? Perhaps I dont understand the question?

Rackover:
No sorry I had answered your message before looking at the new diff so I edited it.
What I meant is that the prefix is set but in case of error it is never cleared. why so?

diamante0018:
I see, well while it would be great to clear, in this case, it would not affect the functionality of the code. Becasue it is set everytime when a command line is specified, whenevr a file is parsed it gets set again with the filename as prefix. In short the functionality is haltered. However if simply for clarity you want me to add a line that also clears it in case of the error, I guess i can do that. It can be confusing but i think it actually does not really matter in this case. However I will push an update

Rackover (approved):

diamante0018:
At last, green CI and A:b::b:roval 🙏

PR #164, opened 2024-11-03
Feat/improve exceptions

SnowyWhite:
**What does this PR do?**

This pull request improves the message shown when the game crashes.  
It replaces the old MessageBox with a [TaskDialog](https://learn.microsoft.com/en-us/windows/win32/api/commctrl/nf-commctrl-taskdialogindirect) that offers more customisation options.

The following information is included in the new message:
* The IW4x revision of the client
* The Windows Version
* The launch parameters

And if the crash happened in-game:
* Whether it was in private match or on a dedicated server
* The IW4x revision of the server
* The server name
* the IP address of the server
* The gametype
* The map name

It also includes a hyperlink to join the Discord server.

To make this possible, a few other things were added or changed too.  

* `sv_version` dvar, which contains the version parsed from an Infostring
* `shortversion` dvar, which contains the client version (was already part of the game)
* `UI_GetGameTypeDisplayName` method, for completeness, not really required
* A new utility method the get the Windows version as a string
* The required minimum version of Windows is updated from 5 to 6 (Vista), as the new Windows Controls library has this limit

<details>

<summary>Screenshot of crash messages</summary>

### Crash happened when playing on a dedicated server

dedicated_server

### Crash happened when playing a private match

private_match

### Crash happened somewhere at the main menu

pregame

</details>


src/Components/Modules/Exception.cpp
```diff
@@ -53,6 +56,144 @@ namespace Components
...
+				Server Info:
+				Type: Dedicated Server							
+				IW4x Version: {}
+				Server Name: {}
+				IP Address: {}					
+				Gametype: {}
+				Map Name: {})",
+				serverVersion, serverName, ipAddress, gameType, mapName);
+
+			std::string msg = std::format("{}\n{}\n\nDo you want to copy this message to the clipboard?", clientInfo, serverInfo);
```
Rackover, line 118: "Do you want to copy this message to the clipboard?" appears three times here - it should be a `constexpr` variable defined earlier, and a part of std::format
```
constexpr auto CLIPBOARD_MSG = "Do you want to copy this message to the clipboard?";

msg = std::format("{}\n{}\n\n{}", clientInfo, serverInfo, CLIPBOARD_MSG);

```

src/Components/Modules/Exception.cpp
```diff
@@ -53,6 +56,144 @@ namespace Components
...
+		::TaskDialogIndirect(&taskDialogConfig, nullptr, nullptr, nullptr);
+	}
+
+	HRESULT CALLBACK Exception::TaskDialogCallbackProc(HWND, UINT notification, WPARAM clickedButton, LPARAM, LONG_PTR data)
+	{
+		const auto* msg = reinterpret_cast<const std::wstring*>(data);
+
+		if (notification == TDN_HYPERLINK_CLICKED)
+		{
+			Utils::OpenUrl("https://discord.gg/2ETE8engZM");
```
Rackover, line 182: Same thing here - this Discord URL link is duplicated, make it either a macro or a constexpr in the class.
The day the discord changes location we will have a single place to change

src/Components/Modules/Exception.cpp
```diff
@@ -98,20 +239,25 @@ namespace Components
 			return EXCEPTION_CONTINUE_EXECUTION;
 		}
 
-		const char* error;
 		if (ExceptionInfo->ExceptionRecord->ExceptionCode == EXCEPTION_STACK_OVERFLOW)
 		{
-			error = "Termination because of a stack overflow.\nCopy exception address to clipboard?";
+			const char* error = "Termination because of a stack overflow.\nCopy exception address to clipboard?";
```
Rackover, line 244: Ditto

src/Utils/Utils.cpp
```diff
@@ -50,6 +50,36 @@ namespace Utils
...
+				const auto minor = versionInfo.dwMinorVersion;
+				const auto build = versionInfo.dwBuildNumber;
+
+				if (major == 10 && build >= 22000) return std::format("Windows 11 (Build {})", build);
+				if (major == 10)				   return std::format("Windows 10 (Build {})", build);
+				if (major == 6 && minor == 3)	   return std::format("Windows 8.1 (Build {})", build);
+				if (major == 6 && minor == 2)	   return std::format("Windows 8.0 (Build {})", build);
+				if (major == 6 && minor == 1)	   return std::format("Windows 7 (Build {})", build);
+				if (major == 6 && minor == 0)	   return std::format("Windows Vista (Build {})", build);
+				if (major == 5 && minor == 1)	   return std::format("Windows XP (Build {})", build);
```
Rackover, line 76: This does not follow the formatting of other code but in this case i'd say that's okay
However if possible i'd like you to add the word in addition to the OS name (32 bit or 64 bit)

Rackover (changes requested):
Looks good overall, few things to change

src/Components/Modules/Exception.cpp
```diff
@@ -9,6 +9,9 @@
 #include <cwctype>
 #include <version.hpp>
 
+#define CLIPBOARD_MSG "Do you want to copy this message to the clipboard?"
+#define DISCORD_LINK  "https://discord.gg/2ETE8engZM"
```
Rackover, line 13: Sorry to be annoying about this but i think you should use constexpr instead of macro for this case (and in general!)
They're more C++ish and give a clear indication to the compiler and to the IDE, about their type for instance, etc
Macros should be avoided in that situation

Rackover:
I'm testing this as soon as I have a bit of time

src/Utils/Utils.cpp
```diff
@@ -50,6 +50,36 @@ namespace Utils
...
+		const auto ntdll = Utils::Library("ntdll.dll");
+		const auto rtlGetVersion = ntdll.getProc<LONG(WINAPI*)(PRTL_OSVERSIONINFOW)>("RtlGetVersion");
+
+		if (rtlGetVersion)
+		{
+			LONG STATUS_SUCCESS = 0;
+			RTL_OSVERSIONINFOW versionInfo = {};
+			versionInfo.dwOSVersionInfoSize = sizeof(versionInfo);
+
+			if (rtlGetVersion(&versionInfo) == STATUS_SUCCESS)
```
wroyca, line 64: Not necessary; it will always succeed. https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlgetversion#return-value

`rtlGetVersion(&versionInfo);` is enough

src/Utils/Utils.cpp
wroyca: If we check for 5.1, we might as well check for 5.2 (Windows XP Professional x64 Edition). https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-osversioninfoexw

src/Utils/Utils.hpp
```diff
@@ -14,6 +14,12 @@ namespace Utils
 
 	bool IsWineEnvironment();
 
+	/// <summary>
```
wroyca, line 17: Don't use [XML documentation](https://learn.microsoft.com/en-us/cpp/build/reference/xml-documentation-visual-cpp?view=msvc-170) :)

src/Components/Modules/Exception.cpp
```diff
@@ -3,11 +3,17 @@
...
+#include "Party.hpp"
+#include "TextRenderer.hpp"
 
+#include <cwctype>
 #include <version.hpp>
 
 namespace Components
 {
+	constexpr auto CLIPBOARD_MSG = "Do you want to copy this message to the clipboard?";
+	constexpr auto DISCORD_LINK = "https://discord.gg/2ETE8engZM";
```
wroyca, line 15: Not sure we want that Discord link here; there's no guarantee whether it remains valid

src/Game/Dvars.hpp
```diff
@@ -132,6 +132,10 @@ namespace Game
 	extern const dvar_t** cg_scoreboardWidth;
 
 	extern const dvar_t** version;
+	/// <summary>
```
wroyca, line 135: Don't use [XML documentation](https://learn.microsoft.com/en-us/cpp/build/reference/xml-documentation-visual-cpp?view=msvc-170) :)

src/Components/Modules/Exception.cpp
```diff
@@ -53,6 +59,148 @@ namespace Components
...
+				return message;
+			}
+		}
+
+		// Get info for a dedicated server
+		{
+			const auto serverVersion = Dvar::Var("sv_version").get<std::string>();
+			const auto ipAddress = Network::Address(*Game::connectedHost).getString();
+
+			char serverName[256]{ 0 };
```
wroyca, line 107: non-constant bound is dangerous

wroyca (changes requested):

src/Components/Modules/Exception.cpp
Rackover: Where else would we put it?
I think it's okay here but feel free to suggest something else

src/Game/Dvars.hpp
Rackover: why so? care to explain? :P

src/Components/Modules/Exception.cpp
Rackover: feel free to accompany your comment with a snippet or a suggestion for something that the offending code should be replaced with

src/Components/Modules/Exception.cpp
```diff
@@ -53,6 +59,148 @@ namespace Components
...
+			std::wstring message(msg.begin(), msg.end());
+			return message;
+		}
+
+		const auto* gameType = (*Game::sv_gametype)->current.string;
+		const auto* mapName = (*Game::sv_mapname)->current.string;
+
+		// Get info for a private match
+		{
+			if (std::strcmp(Game::cls->servername, "localhost") == 0)
```
Rackover, line 87: Somehow i missed this part!
I think you can simply use `Game::cls->servername == "localhost"s` but ....
if you're looking to check whether it's a local server or not there's an easier way: `Game::sv_running`

src/Game/Dvars.hpp
Rackover: yea but as a result of it being microsoft-specific, this comment is visible in Intellisense when you type the name of the function or variable in Visual Studio.

If you type Game::version it will print in a tooltip the content of the <summary> placed just above, making it convenient to use. I do not have the same behaviour with just `///` (although maybe there are other ways to achieve it)

If you have a practical case where the XML Documentation actually poses a problem, one that I can verify, in comparison to slash code documentation (or other methods) tell me! I do not know one, but I also work predominantly with M$ tooling

src/Components/Modules/Exception.cpp
Rackover: Do you have a link for that? (I'm not sure what you mean with "server widget")
The link here is the one supplied on Alterware discord as being permanent (or probably expected to live as long as alterware), is there a better link we should replace this one with?

wroyca: cc @diamante0018 server widget must be enabled explicitly in the server by the owner, https://discord.com/blog/add-the-discord-widget-to-your-site, would you mind taking a look? :)

diamante0018: It's enabled and in use at https://alterware.dev/

Rackover: This spouts yet another discord server link which is neither the one on alterware discord nor the one in the code right there. I'm not sure what that means

src/Components/Modules/Party.cpp
```diff
@@ -582,7 +590,7 @@ namespace Components
 							else
 							{
 								Dvar::Var("xblive_privateserver").set(true);
-
+								ServerVersion.set(version.c_str());
```
diamante0018, line 593: no need for c_str() set takes std::strings as well

src/Game/Dvars.hpp
Rackover: 🤔

If what you mean by " 👎 " is that Visual Studio doesn't handle XML comments in Intellisense, I can provide examples where it does.
If what you mean by " 👎 " is that you know no practical downsides to using XML documentation then i'm happy to mark this conversation as resolved.

If you meant to express anything else, please use your words, I'm not fluent in "👎 "  😄

src/Components/Modules/Exception.cpp
wroyca: So basically, the Discord widget is crafted from the Discord server ID (as opposed to a randomly generated invite link), which is why I believe it's more straightforward, and this can be enabled in the server settings (Diamante says that it is already, but I don't see it?) 

Anyways, it's nit;

src/Game/Dvars.hpp
Rackover: My question is: Is there anything that XML documentation *breaks*  in comparison to `///`? For instance, is there a widely used code editor that simply won't accept or display XML comments but would display `///` comments?

If this is not the case then XML comments are the way to go - a large portion of people who work with this project are using or will use Visual Studio predominantly, where having Intellisense enabled by default means XML comments do have an advantage over `///` comments. But if XML comments breaks other widely used IDEs, I don't mind considering leaving them out of the codebase.

I'm looking for a practical example of XML docs presenting a problem, in a real environment - i understand the theory behind having things standard or not but it's just not something that strikes me as particularly important in that project or in that ecosystem, so practical cases where something actually breaks take priority.

src/Components/Modules/Exception.cpp
wroyca: None that come to mind right now without more context. I believe the first step would be to look into StripColors and StripAllTextIcons and see how we can handle it better; perhaps leave a TODO for now or a FIXME;

src/Game/Dvars.hpp
wroyca: > My question is: Is there anything that XML documentation _breaks_ in comparison to `///`? For instance, is there a widely used code editor that simply won't accept or display XML comments but would display `///` comments?
> 
> If this is not the case then XML comments are the way to go - a large portion of people who work with this project are using or will use Visual Studio predominantly, where having Intellisense enabled by default means XML comments do have an advantage over `///` comments. But if XML comments breaks other widely used IDEs, I don't mind considering leaving them out of the codebase.
> 
> I'm looking for a practical example of XML docs presenting a problem, in a real environment - i understand the theory behind having things standard or not but it's just not something that strikes me as particularly important in that project or in that ecosystem, so practical cases where something actually breaks take priority.

Ah, sorry, I misunderstood your earlier comment. :)

It will be fine – it will just display the tag in the hover doc everywhere else.

e.g. we'll see <summary> foo </summary> as opposed to summary: foo (or something like that).

image

> i understand the theory behind having things standard or not but it's just not something that strikes me as particularly important in that project or in that ecosystem

Agree to disagree; in any case, it's fine to mark it as resolved.

SnowyWhite: I used the XML style comments on purpose for a couple of reasons:

1. As mentioned by louve, one just needs to type `///` and it will automatically create the `<summary>` tag and others for return types and parameters etc.

2. Being able to distinguish them from other comments might come in handy in the future (perhaps the wiki will get a section explaining the codebase, and XML comments would make it easy for automated export and formatting). 

3. I am aware that this is a Visual Studio specific feature, but on the other hand most people will be working on the project with this IDE anyway and will therefore benefit from it, while users of other IDEs will not have much of a disadvantage. Moreover, the XML style is just the default option in Visual Studio, there are other, more standardized options, switching to another that works for everyone would be no problem:
docs

4. Having an eye-catching doc style will hopefully remind and encourage others to document their code too :D

src/Components/Modules/Exception.cpp
SnowyWhite: What I am trying to achieve is to check whether the client was playing a private match (in other words: offline) or on a dedicated server.

While there is no `Game::sv_running` dvar in the code, there is another one called `Game::com_sv_running`.  
Checking this dvar would be essentially the same as checking `Dedicated::IsRunning()`, if I'm not mistaken, which is not what I am looking for.

However, I was thinking about the case where not the client but the server crashes. At the moment, it seems to me that the `Exception` class will handle both scenarios in the same way, and I'm not sure if that's really desired (what happens in a headless environment when a `MessageBox` is presented?).

Rackover: sv_running is a Dvar in IW4 that tells you whether the current instance is running a server or not. If they _are_ the server, then they are not _connected_ to a server - AKA, they're playing on their own machine!
A dedicated is certainly running a server yes, so it has this variable set to true.
I'm not sure if com_sv_running from IW4x is the same thing, i have not had time to check the address - if you can check the Dvar registration of sv_running in IW4 disassembly this will tell you more 😃

src/Components/Modules/Exception.cpp
SnowyWhite: I opened issue #165 for this.

src/Components/Modules/Exception.cpp
SnowyWhite: I did a test with this dvar before I sent my response and it didn't work, but that's because I must have done something wrong. A second test now confirmed that you are absolutely correct 😄 
In IW4, the registered dvar for "sv_running" can be found at `0x1AD7934`, which happens to be `com_sv_running` in IW4x.
Should I rename it? There are probably more wrongly named dvars 👀 
image

Rackover: Check that there's not a shenanigan in the engine, like the command registered twice in the same dvar with different names
If they're none, then yes, this ought to be renamed sv_running 😄

Rackover:
So it's really cool and it passes testing but it's missing one thing..... `fs_game` 🤔 
image

I think this is just an oversight cause I remember you talking about it.
Here I'm on a heavily modded server, fs_game is definitely not vanilla, so it should probably be part of the exception message!

Once this is handled i'm ready to merge it & release

Rackover (approved):

PR #166, opened 2024-11-09
Fix of fps drop when using a high polling rate mouse

L3D451R7:
This pull request fixes an fps drop caused by bottleneck of window messages when using high polling rate mouse (8k).

These changes includes RIDEV_NOLEGACY flag to RegisterRawInputDevices, disabling legacy WindowProc messages for input device we register.

I had to fix issues with mouse input when m_raw_input is 0, and also issues that are including with lack of legacy messages 
that i managed to fix with disabling raw input (handling WM_KILLFOCUS, WM_SETFOCUS, mouse going away from client screen space (WM_MOUSELEAVE)).

Since the legacy messages are gone, i made mouse input fully raw when using m_raw_input 1 (handling buttons, scroll).

here is the result (m_raw_input 0 will create fps drops since it uses game's old behaviour):

https://github.com/user-attachments/assets/40b69069-132b-409f-8538-f48cfaeef68a

### DIFF
--- a/src/Components/Modules/RawMouse.cpp
+++ b/src/Components/Modules/RawMouse.cpp
@@ -9,18 +9,16 @@ namespace Components
 	int RawMouse::MouseRawX = 0;
 	int RawMouse::MouseRawY = 0;
 	bool RawMouse::InRawInput = false;
+	bool RawMouse::RawInputWasEverEnabled = false;
 
 #define K_MWHEELUP 205
 #define K_MWHEELDOWN 206
 
 #define HIGH_POLLING_FIX 1
 
-	void RawMouse::IN_ClampMouseMove()
+	void ClampMousePos(POINT& curPos)
 	{
 		tagRECT rc;
-		tagPOINT curPos;
-
-		GetCursorPos(&curPos);
 		GetWindowRect(Window::GetWindow(), &rc);
 		auto isClamped = false;
 		if (curPos.x >= rc.left)
@@ -36,6 +34,7 @@ namespace Components
 			curPos.x = rc.left;
 			isClamped = true;
 		}
+
 		if (curPos.y >= rc.top)
 		{
 			if (curPos.y >= rc.bottom)
@@ -56,8 +55,18 @@ namespace Components
 		}
 	}
 
+	void RawMouse::IN_ClampMouseMove()
+	{
+		tagPOINT curPos;
+		GetCursorPos(&curPos);
+		ClampMousePos(curPos);
+	}
+
 	BOOL RawMouse::OnRawInput(LPARAM lParam, WPARAM)
 	{
+		if (!M_RawInput.get<bool>())
+			return TRUE;
+
 		auto dwSize = sizeof(RAWINPUT);
 		static BYTE lpb[sizeof(RAWINPUT)];
 
@@ -144,8 +153,6 @@ namespace Components
 
 	void RawMouse::IN_RawMouseMove()
 	{
-		static auto r_fullscreen = Dvar::Var("r_fullscreen");
-
 		if (GetForegroundWindow() != Window::GetWindow())
 			return;
 
@@ -180,7 +187,7 @@ namespace Components
 	{
 		if (!M_RawInput.get<bool>())
 		{
-			if (!InRawInput)
+			if (!InRawInput || !RawInputWasEverEnabled)
 				return false;
 
 			enable = false;
@@ -192,18 +199,21 @@ namespace Components
 		}
 
 #if HIGH_POLLING_FIX
-		constexpr DWORD rawinput_flags = RIDEV_INPUTSINK | RIDEV_NOLEGACY | RIDEV_CAPTUREMOUSE;
+		constexpr DWORD rawinput_flags = RIDEV_INPUTSINK | RIDEV_NOLEGACY;
 #else
 		constexpr DWORD rawinput_flags = RIDEV_INPUTSINK;
 #endif
 
 		RAWINPUTDEVICE Rid[1];
-		Rid[0].usUsagePage = 0x01; // HID_USAGE_PAGE_GENERIC
-		Rid[0].usUsage = 0x02; // HID_USAGE_GENERIC_MOUSE
+		Rid[0].usUsagePage = HID_USAGE_PAGE_GENERIC;
+		Rid[0].usUsage = HID_USAGE_GENERIC_MOUSE;
 		Rid[0].dwFlags = (enable ? rawinput_flags : RIDEV_REMOVE);
 		Rid[0].hwndTarget = enable ? Window::GetWindow() : NULL;
 
 		bool success = RegisterRawInputDevices(Rid, ARRAYSIZE(Rid), sizeof(Rid[0])) == TRUE;
+
+		if (success && enable) //
+			RawInputWasEverEnabled = true;
 
 		InRawInput = success && (Rid[0].dwFlags & RIDEV_REMOVE) == 0;
 #ifdef _DEBUG
@@ -213,13 +223,12 @@ namespace Components
 			Logger::Debug("Raw Mouse disabled");
 #endif
 		return true;
-}
+	}
 
 	void RawMouse::IN_RawMouse_Init()
 	{
 		if (Window::GetWindow() && ToggleRawInput(true)) {
 			Logger::Debug("Raw Mouse Init");
-			InRawInput = true;
 		}
 	}
 
@@ -252,13 +261,42 @@ namespace Components
 
 	void RawMouse::IN_MouseMove()
 	{
+		static auto r_fullscreen = Dvar::Var("r_fullscreen");
+
 		if (InRawInput)
 		{
-			IN_RawMouseMove();
+			return IN_RawMouseMove();
 		}
-		else
+
+		// IN_MouseMove
+		if (GetForegroundWindow() != Window::GetWindow())
+			return;
+
+		tagPOINT cursor_pos;
+		static tagPOINT prev_cursor;
+
+		GetCursorPos(&cursor_pos);
+		if (r_fullscreen.get<bool>())
+			ClampMousePos(cursor_pos);
+
+		int delta_x = cursor_pos.x - prev_cursor.x;
+		int delta_y = cursor_pos.y - prev_cursor.y;
+		prev_cursor = cursor_pos;
+
+		ScreenToClient(Window::GetWindow(), &cursor_pos);
+		auto recenterMouse = Game::CL_MouseEvent(cursor_pos.x, cursor_pos.y, delta_x, delta_y);
+
+		if (recenterMouse && (delta_x || delta_y))
 		{
-			Game::IN_MouseMove();
+			RECT Rect;
+			GetWindowRect(Window::GetWindow(), &Rect);
+
+			int window_center_x = (Rect.right + Rect.left) / 2;
+			int window_center_y = (Rect.top + Rect.bottom) / 2;
+			SetCursorPos(window_center_x, window_center_y);
+
+			prev_cursor.x = window_center_x;
+			prev_cursor.y = window_center_y;
 		}
 	}
 
@@ -266,22 +304,28 @@ namespace Components
 	{
 		Utils::Hook(0x475E65, IN_MouseMove, HOOK_JUMP).install()->quick();
 		Utils::Hook(0x475E8D, IN_MouseMove, HOOK_JUMP).install()->quick();
+		Utils::Hook(0x475E9E, IN_MouseMove, HOOK_JUMP).install()->quick();
 
 		Utils::Hook(0x467C03, IN_Init, HOOK_CALL).install()->quick();
 		Utils::Hook(0x64D095, IN_Init, HOOK_JUMP).install()->quick();
 
+#if HIGH_POLLING_FIX
 		Utils::Hook(0x60BFB9, IN_Frame, HOOK_CALL).install()->quick();
 		Utils::Hook(0x4A87E2, IN_Frame, HOOK_CALL).install()->quick();
 		Utils::Hook(0x48A0E6, IN_Frame, HOOK_CALL).install()->quick();
+#endif
 
 		Utils::Hook(0x473517, IN_RecenterMouse, HOOK_CALL).install()->quick();
 		Utils::Hook(0x64C520, IN_RecenterMouse, HOOK_CALL).install()->quick();
 
 		M_RawInput = Dvar::Register<bool>("m_rawinput", true, Game::DVAR_ARCHIVE, "Use raw mouse input, Improves accuracy & has better support for higher polling rates");
 
-		Window::OnWndMessage(WM_INPUT, OnRawInput);
+#if HIGH_POLLING_FIX
 		Window::OnWndMessage(WM_KILLFOCUS, OnKillFocus);
 		Window::OnWndMessage(WM_SETFOCUS, OnSetFocus);
+#endif
+
+		Window::OnWndMessage(WM_INPUT, OnRawInput);
 		Window::OnCreate(IN_RawMouse_Init);
 	}
 }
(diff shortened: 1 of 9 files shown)


src/Components/Modules/RawMouse.cpp
```diff
@@ -50,79 +55,180 @@ namespace Components
...
+
+		int mouse_event_flag = (raw->data.mouse.usButtonFlags & RI_MOUSE_BUTTON_1_DOWN) != 0;
+		if ((raw->data.mouse.usButtonFlags & RI_MOUSE_BUTTON_2_DOWN) != 0)
+			mouse_event_flag |= 2u;
+		if ((raw->data.mouse.usButtonFlags & RI_MOUSE_BUTTON_3_DOWN) != 0)
+			mouse_event_flag |= 4u;
+		if ((raw->data.mouse.usButtonFlags & RI_MOUSE_BUTTON_4_DOWN) != 0)
+			mouse_event_flag |= 8u;
+		if ((raw->data.mouse.usButtonFlags & RI_MOUSE_BUTTON_5_DOWN) != 0)
+			mouse_event_flag |= 0x10u;
```
wroyca, line 103: don't we need to check for up as well? e.g. 'raw->data.mouse.usButtonFlags & (RI_MOUSE_BUTTON_5_DOWN | RI_MOUSE_BUTTON_5_UP)) != 0'

wroyca (changes requested):

src/Components/Modules/RawMouse.cpp
L3D451R7: yeah, fixed it now

src/Components/Modules/RawMouse.cpp
```diff
@@ -8,13 +8,17 @@ namespace Components
 	Dvar::Var RawMouse::M_RawInput;
 	int RawMouse::MouseRawX = 0;
 	int RawMouse::MouseRawY = 0;
+	bool RawMouse::InRawInput = false;
+	bool RawMouse::RawInputWasEverEnabled = false;
 
-	void RawMouse::IN_ClampMouseMove()
+#define K_MWHEELUP 205
```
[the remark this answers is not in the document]
L3D451R7: done

src/Components/Modules/RawMouse.cpp
```diff
@@ -50,79 +55,180 @@ namespace Components
...
+		constexpr DWORD rawinput_flags = RIDEV_INPUTSINK;
+#endif
+
+		RAWINPUTDEVICE Rid[1];
+		Rid[0].usUsagePage = HID_USAGE_PAGE_GENERIC;
+		Rid[0].usUsage = HID_USAGE_GENERIC_MOUSE;
+		Rid[0].dwFlags = (enable ? rawinput_flags : RIDEV_REMOVE);
+		Rid[0].hwndTarget = enable ? Window::GetWindow() : NULL;
+
+		bool success = RegisterRawInputDevices(Rid, ARRAYSIZE(Rid), sizeof(Rid[0])) == TRUE;
```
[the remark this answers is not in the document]
L3D451R7: added check for raw input being present.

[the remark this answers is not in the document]
L3D451R7: added error logging in case of failure.

src/Components/Modules/RawMouse.cpp
```diff
@@ -50,86 +56,233 @@ namespace Components
...
+
 	BOOL RawMouse::OnRawInput(LPARAM lParam, WPARAM)
 	{
+		if (!M_RawInput.get<bool>()) {
+			ResetMouseRawEvents();
+			return TRUE;
+		}
+
 		auto dwSize = sizeof(RAWINPUT);
 		static BYTE lpb[sizeof(RAWINPUT)];
```
Rackover, line 92: This probably ought to be simply `static RAWINPUT lpb;` - this way the size is correct and no need for Reinterpret_cast below

src/Components/Modules/RawMouse.cpp
```diff
@@ -50,86 +56,233 @@ namespace Components
...
 
 		GetRawInputData((HRAWINPUT)lParam, RID_INPUT, lpb, &dwSize, sizeof(RAWINPUTHEADER));
 
 		auto* raw = reinterpret_cast<RAWINPUT*>(lpb);
```
Rackover, line 96: `raw = &lpb;` (provided lpb is made the correct type)

src/Components/Modules/RawMouse.cpp
```diff
@@ -50,86 +56,233 @@ namespace Components
...
 	{
+		if (!M_RawInput.get<bool>()) {
+			ResetMouseRawEvents();
+			return TRUE;
+		}
+
 		auto dwSize = sizeof(RAWINPUT);
 		static BYTE lpb[sizeof(RAWINPUT)];
 
 		GetRawInputData((HRAWINPUT)lParam, RID_INPUT, lpb, &dwSize, sizeof(RAWINPUTHEADER));
```
Rackover, line 94: - Return value never handled (it can return -1 in case of error, or otherwise) - best to check that to avoid processing invalid mouse inputs
- `static_cast<>` or `reinterpret_cast<>` are preferred to C-Style casts `()`

src/Components/Modules/RawMouse.cpp
```diff
@@ -50,86 +56,233 @@ namespace Components
...
 	{
-		if (Window::GetWindow() && M_RawInput.get<bool>())
+		RawInputSupported = false;
+
+		HMODULE user32 = GetModuleHandleA("USER32.dll");
+		if (user32)
 		{
-			Logger::Debug("Raw Mouse Init");
+			PVOID pfnRegisterRawInputDevices = GetProcAddress(user32, "RegisterRawInputDevices");
+			PVOID pfnGetRawInputData = GetProcAddress(user32, "GetRawInputData");
```
Rackover, line 261: These two are Windows XP functions so you can access them directl ywithout using GetModuleHandleA and GetProcAddress I think !

src/Components/Modules/RawMouse.cpp
```diff
@@ -146,29 +299,70 @@ namespace Components
 
 	void RawMouse::IN_MouseMove()
 	{
-		if (M_RawInput.get<bool>())
+		static auto r_fullscreen = Dvar::Var("r_fullscreen");
```
Rackover, line 302: r_fullscreen is registered at R_Init, which is very early, so rather than declaring the Dvar (like you do for autopriority) I think you'd rather add this dvar to `Game::` and reference it directly.
It is directly access at the beginning of IN_MouseMove in the engine so I assume this is supposed to exist at that time

The address is 0x069F0DA0

src/Components/Modules/RawMouse.cpp
```diff
@@ -146,29 +299,70 @@ namespace Components
...
+		prev_cursor = cursor_pos;
+
+		ScreenToClient(Window::GetWindow(), &cursor_pos);
+		auto recenterMouse = Game::CL_MouseEvent(cursor_pos.x, cursor_pos.y, delta_x, delta_y);
+
+		if (recenterMouse && (delta_x || delta_y))
 		{
-			Game::IN_MouseMove();
+			RECT Rect;
+			GetWindowRect(Window::GetWindow(), &Rect);
```
Rackover, line 330: Believe it or not, GetWindowRect can fail (check for == TRUE before doing anything with RECT, otherwise you're using unitialized struct)

src/Components/Modules/RawMouse.cpp
```diff
@@ -146,29 +299,70 @@ namespace Components
...
 		Utils::Hook(0x475E8D, IN_MouseMove, HOOK_JUMP).install()->quick();
+		Utils::Hook(0x475E9E, IN_MouseMove, HOOK_JUMP).install()->quick();
 
 		Utils::Hook(0x467C03, IN_Init, HOOK_CALL).install()->quick();
 		Utils::Hook(0x64D095, IN_Init, HOOK_JUMP).install()->quick();
 
+#if HIGH_POLLING_RATE_FIX
+		Utils::Hook(0x60BFB9, IN_Frame, HOOK_CALL).install()->quick();
+		Utils::Hook(0x4A87E2, IN_Frame, HOOK_CALL).install()->quick();
+		Utils::Hook(0x48A0E6, IN_Frame, HOOK_CALL).install()->quick();
```
Rackover, line 353: Random bit of review, feel free to mark this as resolved if due dilligence has been taken:
Please make sure nowhere in the IW4x client has these 3 addresses already hooked. Gamepad comes to mind as a module that hooks into INput frame in several places, so please make sure this is not disrpting another module (make sure gamepad is still usable and make sure these addresses are not mentioned elsewhere in the source code)

src/Game/System.cpp
```diff
@@ -27,6 +27,20 @@ namespace Game
 	Sys_OutOfMemErrorInternal_t Sys_OutOfMemErrorInternal = Sys_OutOfMemErrorInternal_t(0x4B2E60);
 	Sys_QuitAndStartProcess_t Sys_QuitAndStartProcess = Sys_QuitAndStartProcess_t(0x45FCF0);
 
+	Sys_QueEvent_t Sys_QueEvent = Sys_QueEvent_t(0x497DE0);
+
+	int &sysMsgTime()
+	{
+		static const DWORD sysMsgTime_t = 0x64A3AE8;
+		return *(int*)sysMsgTime_t;
```
Rackover, line 35: This is a little bit ugly and hard to parse: 
- 0x64A3AC8 is the address of a variable named `g_wv` of type `WinVars_t`. This is probably the object you want to access instead of fetching the member sysMsgTime directly
- _t is a suffix for type and i'm not sure it's appropriate to name a variable like this
- C++ Cast (reinterpret/static) is preferred to C-Style (*)
- Wrong casing on the function name (?)

Probably a better option here is to properly hook g_wv (WinVars_t*) by its address, as a public member of Game::, then letting your other code access that variable's member directly (or through a function)

Rackover (changes requested):
Sorry to point out stuff that may not be coming from _your_ PR in particular, but It's also my first look at `RawMouse.cpp` so while we're here I figure we can
Other than that there are a few things that need adjustment and then once this passes building in CI i'll test it before merging 😃

src/Components/Modules/RawMouse.cpp
L3D451R7: yeah, thats just a safety check that can be left out

src/Components/Modules/RawMouse.cpp
L3D451R7: just rechecked those, all of them have only 1 reference in RawMouse.cpp

src/Game/System.cpp
L3D451R7: done

src/Components/Modules/RawMouse.cpp
L3D451R7: done

src/Components/Modules/RawMouse.cpp
L3D451R7: done, now it initializes all the DVars in IN_Init, and checks if DVar was inited - you can remove the initialization in IN_Init and code will not crash cause of that.

src/Components/Modules/RawMouse.cpp
L3D451R7: done

src/Components/Modules/RawMouse.cpp
L3D451R7: done

src/Components/Modules/RawMouse.cpp
L3D451R7: done

src/Components/Modules/RawMouse.cpp
L3D451R7: just rechecked, MW2 (and iw4x) does not support windows older than XP - so yeah this check is completely useless.

src/Components/Modules/RawMouse.cpp
```diff
@@ -146,29 +358,69 @@ namespace Components
...
+			return;
+
+		tagPOINT cursor_pos;
+		static tagPOINT prev_cursor;
+
+		GetCursorPos(&cursor_pos);
+		if (R_FullScreen.get<Game::dvar_t*>() && R_FullScreen.get<bool>())
+			ClampMousePos(cursor_pos);
+
+		int delta_x = cursor_pos.x - prev_cursor.x;
```
Rackover, line 377: iw4x is camelCase  !

src/Components/Modules/RawMouse.cpp
```diff
@@ -146,29 +358,69 @@ namespace Components
...
-			Game::IN_MouseMove();
+			RECT Rect;
+			if (GetWindowRect(Window::GetWindow(), &Rect) == TRUE)
+			{
+				int window_center_x = (Rect.right + Rect.left) / 2;
+				int window_center_y = (Rect.top + Rect.bottom) / 2;
+				SetCursorPos(window_center_x, window_center_y);
+
+				prev_cursor.x = window_center_x;
+				prev_cursor.y = window_center_y;
```
Rackover, line 394: camelCase

src/Components/Modules/RawMouse.cpp
```diff
@@ -3,19 +3,48 @@
 #include "RawMouse.hpp"
 #include "Window.hpp"
 
+#define HIGH_POLLING_RATE_FIX 1
+#define RAW_MOUSE_VERBOSE 1
+
 namespace Components
 {
+	void raw_mouse_value_t::ResetDelta()
```
Rackover, line 11: Please prefer camelCase instead of snake_case, for type names too
The only exception is for dvar names and game structs wheree we try to stick to the game's convention (or lack thereof)

src/Components/Modules/RawMouse.cpp
```diff
@@ -3,19 +3,48 @@
 #include "RawMouse.hpp"
 #include "Window.hpp"
 
+#define HIGH_POLLING_RATE_FIX 1
+#define RAW_MOUSE_VERBOSE 1
```
Rackover, line 7: I suppose I understand the rationale behind having HIGH_POLLING_RATE_FIX as macro (is there ever a case where we _don't want_ the fix though?)

However Raw mouse verbose sounds like it should be a Dvar instead. Turning on or off mouse log should not require to recompile the software 🫡 

Please make it a Dvar if possible (with ::DVAR_SAVED flag maybye?)

src/Components/Modules/RawMouse.cpp
```diff
@@ -50,86 +80,268 @@ namespace Components
...
 	}
 
 	void RawMouse::IN_Init()
 	{
 		Game::IN_Init();
 		IN_RawMouse_Init();
+		ResetMouseRawEvents();
+
+		R_AutoPriority = Dvar::Var("r_autopriority");
+		R_FullScreen = Dvar::Var(0x069F0DA0);
```
Rackover, line 335: nice, I didn't know about this constructor!

Rackover (changes requested):
Only very small stuff remains

Rackover (changes requested):
After testing this makes gamepad **non fonctional** compared to r4695
- Gamepad cannot be used to navigate in menus
- Gamepad is not recognized ingame either

I think the hooks you have made must conflict with something within gamepad support. Maybe they overwrite or something or are two wide. Try disabling the hooks you added and see if gamepad works again - reenable the most suspicious hook, and so on until stumbling on the one that breaks it

src/Components/Modules/RawMouse.cpp
L3D451R7: done.

src/Components/Modules/RawMouse.cpp
L3D451R7: done.

src/Components/Modules/RawMouse.cpp
L3D451R7: done.

Rackover (approved):

Rackover:
Tested, no undesirable side effects (mouse too shitty to test the high DPI)

PR #168, opened 2024-11-29
fixed codol LOD

nucleur-goomer:
**What does this PR do?**

This PR fixes the LOD issues on CoDOL maps

The LODs which formerly looked liked mashed up polygons now resemble the models of which they truly are.
Ex: Trees on Fav & Estate Tropical, Barrels on Rust Long, etc

Idk, it's one line.

yes
Not an open issue but a well known one
It's a copy paste of someone else's code with the vars in memcpy changed.
one commit


Rackover:
Rust long and Storm spring still look cooked, so I suppose there's more to this than this PR and that we'll need to come back to it at a later point
image
image
image


But LODs on Fav Tropical and Estate Tropical look much better 😃 
image
image

I think it would be really cool if we could replace all those `+ 12` offsets and so on by a real struct description of the rawfile model format, would probably help with this kind of oversight

Rackover (approved):

nucleur-goomer:
Yea the cars are a problem also on bog...  I too am curious about it.  Guess I'll just printf all the stuff on modifyasset and compare iw4 modelsurfs to OL ones again.  I assumed the data structure is a bit different to iw4 so the offset is coping with that.

PR #169, opened 2024-12-02
AntiLag rework

L3D451R7:
This PR entirely reworks existing AntiLag, adding optimized mode and debug logs.

AntiLag was rebuilt, it was debugged and some of the issues were fixed with optimized mode.
I also added back IW3 behavior of AntiLag when its disabled adding optimization too.

As bug example - the throwables were calling AntiLag's rewind even after they stuck in/hit the ground, I decided to keep that optimization with **g_antilag 1** since it's obviously a bug.

With optimized mode (**g_antilag 2**) I also disable AntiLag's rewind on teammates. (I did not find scr_FriendlyFire there so its just calling ingame function)

Also I added attacker in arguments of G_AntiLagRewindClientPos which required rebuilding Weapon_Melee, Bullet_Fire and G_RunMissile.

Throwables issue visualization:

https://github.com/user-attachments/assets/d3357ca8-3cd5-46ca-8ecd-ac0e19ba72c8

After fix:

https://github.com/user-attachments/assets/ec720e49-a577-4cc0-be91-3fae7524425d

### DIFF
--- a/src/Components/Modules/AntiLag.cpp
+++ b/src/Components/Modules/AntiLag.cpp
@@ -62,7 +62,7 @@ namespace Components
 
 		for (int i = 0; i < Game::level->maxclients; ++i)
 		{
-			auto& ent = Game::g_entities[i];
+			Game::gentity_s& ent = Game::g_entities[i];
 
 			bool clientMoved = clientsMoved[i];
 			if (antiLagMode == 2)
@@ -121,7 +121,7 @@ namespace Components
 
 		for (int i = 0; i < Game::level->maxclients; ++i)
 		{
-			auto& ent = Game::g_entities[i];
+			Game::gentity_s& ent = Game::g_entities[i];
 
 			if (!antilagStore->clientMoved[i]) {
 				continue;
--- a/src/Components/Modules/AntiLag.cpp
+++ b/src/Components/Modules/AntiLag.cpp
@@ -189,7 +189,12 @@ namespace Components
 			Logger::Debug("G_RunMissile: at tick {}, level.time {}\n", Game::level->time + throwingDelay, Game::level->time);
 		}
 
-		AntiLag::G_AntiLagRewindClientPos(missile, Game::level->time + throwingDelay, &antilagStore);
+		uint16_t ownerId = missile->r.ownerNum.number;
+		Game::gentity_s* attacker = nullptr;
+		if (ownerId > 0 && ownerId <= Game::level->maxclients) //
+			attacker = &Game::g_entities[static_cast<int>(ownerId) - 1];
+
+		AntiLag::G_AntiLagRewindClientPos(attacker, Game::level->time + throwingDelay, &antilagStore);
 		Game::G_RunMissileInternal(missile);
 		AntiLag::G_AntiLag_RestoreClientPos(&antilagStore);
 	}
--- a/src/Components/Modules/AntiLag.cpp
+++ b/src/Components/Modules/AntiLag.cpp
@@ -191,8 +191,9 @@ namespace Components
 
 		uint16_t ownerId = missile->r.ownerNum.number;
 		Game::gentity_s* attacker = nullptr;
-		if (ownerId > 0 && ownerId <= Game::level->maxclients) //
+		if (ownerId > 0 && ownerId <= Game::level->maxclients) {
 			attacker = &Game::g_entities[static_cast<int>(ownerId) - 1];
+		}
 
 		AntiLag::G_AntiLagRewindClientPos(attacker, Game::level->time + throwingDelay, &antilagStore);
 		Game::G_RunMissileInternal(missile);
--- a/src/Components/Modules/AntiLag.cpp
+++ b/src/Components/Modules/AntiLag.cpp
@@ -6,40 +6,43 @@ namespace Components
 	Dvar::Var AntiLag::G_AntiLag;
 	Dvar::Var AntiLag::G_AntiLagDebug;
 	bool AntiLag::AntiLagFilled = false;
+	float AntiLag::ColorRed[] = { 1.0f, 0.0f, 0.0f, 1.0f };
+	float AntiLag::ColorGreen[] = { 0.0f, 1.0f, 0.0f, 1.0f };
+	float AntiLag::ColorWhite[] = { 1.0f, 1.0f, 1.0f, 1.0f };
 
-	void __cdecl AntiLag::FireWeapon(Game::gentity_s* ent, int gameTime, int eventType)
+	void AntiLag::FireWeapon(Game::gentity_s* ent, int gameTime, int eventType)
 	{
 		if (G_AntiLag.get<int>() == 0)
 			gameTime = Game::level->time;
 		else if (G_AntiLagDebug.get<int>() != 0)
 			Logger::Debug("FireWeapon: at tick {}, level.time {}\n", gameTime, Game::level->time);
 
-		return Game::FireWeapon(ent, gameTime, eventType);
+		Game::FireWeapon(ent, gameTime, eventType);
 	}
 
-	void __cdecl AntiLag::FireWeaponMelee(Game::gentity_s* ent, int gameTime)
+	void AntiLag::FireWeaponMelee(Game::gentity_s* ent, int gameTime)
 	{
 		if (G_AntiLag.get<int>() == 0)
 			gameTime = Game::level->time;
 		else if (G_AntiLagDebug.get<int>() != 0)
 			Logger::Debug("FireWeaponMelee: at tick {}, level.time {}\n", gameTime, Game::level->time);
 
-		return Game::FireWeaponMelee(ent, gameTime);
+		Game::FireWeaponMelee(ent, gameTime);
 	}
 
-	bool AntiLag::ShouldRewind(uintptr_t stack, uintptr_t returnAddress, Game::gentity_s*& attacker)
+	bool AntiLag::ShouldRewind(uintptr_t stack, uintptr_t returnAddress, Game::gentity_s** attacker)
 	{
 		constexpr uintptr_t Bullet_FireRet = 0x004402E0;
 		constexpr uintptr_t Run_MeleeRet = 0x00501A07;
 		constexpr uintptr_t G_RunMissileRet = 0x004AAD46;
 
 		if (Bullet_FireRet == returnAddress) {
-			attacker = *reinterpret_cast<Game::gentity_s**>(stack + 0x14);
+			*attacker = *reinterpret_cast<Game::gentity_s**>(stack + 0x14);
 			return true;
 		}
 
 		if (Run_MeleeRet == returnAddress) {
-			attacker = *reinterpret_cast<Game::gentity_s**>(stack + 0x10);
+			*attacker = *reinterpret_cast<Game::gentity_s**>(stack + 0x10);
 			return true;
 		}
 
@@ -50,7 +53,7 @@ namespace Components
 			return false;
 		}
 
-		auto projectile = *reinterpret_cast<Game::gentity_s**>(stack + 0x10);
+		Game::gentity_s* projectile = *reinterpret_cast<Game::gentity_s**>(stack + 0x10);
 		if (!projectile)
 		{
 			// no debug check since this is totally unexpected to happen at all.
@@ -60,10 +63,10 @@ namespace Components
 
 		int ownerId = projectile->r.ownerNum.number;
 		if (ownerId > 0 && ownerId <= Game::level->maxclients) //
-			attacker = &Game::g_entities[ownerId - 1];
+			*attacker = &Game::g_entities[ownerId - 1];
 
 		const bool hitGround = projectile->s.groundEntityNum == 0x7FE;
-		auto WeaponDef = Game::BG_GetWeaponDef(projectile->s.weapon);
+		Game::WeaponDef* WeaponDef = Game::BG_GetWeaponDef(projectile->s.weapon);
 
 		// l3d note: this check below does not exist in original iw engine.
 		if (hitGround && (WeaponDef->stickiness == Game::WEAPSTICKINESS_ALL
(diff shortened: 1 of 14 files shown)


src/Components/Modules/AntiLag.cpp
```diff
@@ -0,0 +1,216 @@
...
+	{
+		if (G_AntiLag.get<int>() == 0)
+			gameTime = Game::level->time;
+		else if (G_AntiLagDebug.get<int>() != 0)
+			Logger::Debug("FireWeaponMelee: at tick {}, level.time {}\n", gameTime, Game::level->time);
+
+		return Game::FireWeaponMelee(ent, gameTime);
+	}
+
+	bool AntiLag::ShouldRewind(uintptr_t stack, uintptr_t returnAddress, Game::gentity_s*& attacker)
```
diamante0018, line 30: This is not how we do hooks in this client, I have to admit I find this a big hack, it's not proper. I wish to not have code in the client that does hacks such as checking stack and return addresses additionally also the third parameter is a hack. it's should always just be defined as a pointer and not *&, please see other parts of the clients where entity pointers are used.

Hooks should be done following already established practises in the code base with either hook calls, jumps, and asm stubs that call other functions... Please see other modules for more proper examples

Do not define function addresses in code, use the proper files (like functions.hpp or cpp) in the game folder

src/Components/Modules/AntiLag.cpp
```diff
@@ -0,0 +1,216 @@
...
+			|| WeaponDef->stickiness == Game::WEAPSTICKINESS_KNIFE))
+		{
+			// we did stick, no need for rewinding anymore.
+			return false;
+		}
+
+		if (G_AntiLagDebug.get<int>() == 3)
+		{
+			// draw flying projectiles origin
+			std::array<float, 4> clr = { 1.f, 1.f, 1.f, 1.f };
```
diamante0018, line 80: these should probably be static and/or defined as constexpr i think

src/Game/Functions.hpp
```diff
@@ -299,6 +299,18 @@ namespace Game
 	typedef void(*NET_DeferPacketToClient_t)(netadr_t*, msg_t*);
 	extern NET_DeferPacketToClient_t NET_DeferPacketToClient;
 
+	typedef bool(*OnSameTeam_t)(gentity_s* a1, gentity_s* a2);
+	extern OnSameTeam_t OnSameTeam;
+
+	typedef void (*FireWeapon_t)(Game::gentity_s* ent, int gameTime, int a3);
```
diamante0018, line 305: You could leave a space between the two lines block definitions, it's just easier to read and same with the code below

src/Game/Structs.hpp
```diff
@@ -11760,6 +11760,16 @@ namespace Game
 		MssStreamReadInfo streamReadInfo[12];
 	};
 
+	struct AntilagClientStore
```
diamante0018, line 11763: This struct does not belong here, custom definitions of structs with methods would probably belong to the antilag header

diamante0018 (changes requested):
Here are a few things that can be improved in the code

remove `__cdecl` from function definitions

remove return statements on `void` functions

See other comments

src/Components/Modules/AntiLag.cpp
L3D451R7: I can agree to some of your words, but not gonna lie I find really weird that you want to keep those practise with using an asm stub in code that can be done in more readable way

I thought it was the best solution since the other way would require to add more code, or making the code less easily readable when doing an asm stub.

mxve:
Played a few rounds, 2 players on a dedicated server with g_antilag set to 2. Doesn't seem to break anything, at least at low ping.

src/Components/Modules/AntiLag.cpp
```diff
@@ -0,0 +1,219 @@
...
+		else if (G_AntiLagDebug.get<int>() != 0)
+			Logger::Debug("FireWeapon: at tick {}, level.time {}\n", gameTime, Game::level->time);
+
+		Game::FireWeapon(ent, gameTime, eventType);
+	}
+
+	void AntiLag::FireWeaponMelee(Game::gentity_s* ent, int gameTime)
+	{
+		if (G_AntiLag.get<int>() == 0)
+			gameTime = Game::level->time;
```
Rackover, line 26: It's something I noticed on your other PR too - sorry to be annoying with this - but I think I would prefer if we used `{` and `}` systematically, even on "one line" ifs. The reason I say that is because , aesthetics aside, we had a very nasty bug some time ago that was due to something like that (somebody adding one more condition under an existing one, not realizing it didn't have brackets) which was annoying to track
It made me realize it's probably not a gain of time to not use brackets, and we should probably always use them even if the condition block contains a single line.
There are exceptions (switch cases, very long chain of conditions that look similar, ...) but for the general case I think i'd prefer brackets if that's OK

src/Components/Modules/AntiLag.cpp
```diff
@@ -0,0 +1,219 @@
...
+		else if (G_AntiLagDebug.get<int>() != 0)
+			Logger::Debug("FireWeaponMelee: at tick {}, level.time {}\n", gameTime, Game::level->time);
+
+		Game::FireWeaponMelee(ent, gameTime);
+	}
+
+	bool AntiLag::ShouldRewind(uintptr_t stack, uintptr_t returnAddress, Game::gentity_s** attacker)
```
Rackover, line 33: I'm gonna side with Diamante on this, it's a bit nasty to have a C funcution with uintptr_t stack and return address and have it perform operations like that instead of a ASM stub - as in, it's highly unusual here.
I'd rather not have the C code perform low level stuff like this and that this is broken down into an ASM stub + a C/C++ function. 
I'll look further into this later so I can give you more precise suggestions if you need it

src/Components/Modules/AntiLag.cpp
```diff
@@ -0,0 +1,219 @@
...
+		Game::gentity_s* attacker = nullptr;
+
+		// get attacker, check if we should run.
+		if (!ShouldRewind(reinterpret_cast<uintptr_t>(_AddressOfReturnAddress()), reinterpret_cast<uintptr_t>(_ReturnAddress()), &attacker))
+			return;
+
+		antilagStore->Reset();
+
+		// GetClientPositionsAtTime results arrays.
+		std::array<bool, 18> clientsMoved;
```
Rackover, line 109: 18 out to be MAX_PLAYERS constantt

src/Components/Modules/QuickPatch.cpp
```diff
@@ -54,62 +54,6 @@ namespace Components
...
-			mov eax, 0x4A4D50 // FireWeapon
-			call eax
-			add esp, 0Ch
-			pop edi
-			pop ecx
-			retn
-		}
-	}
-
-	__declspec(naked) void QuickPatch::ClientEventsFireWeaponMelee_Stub()
```
Rackover, line 86: Very happy we moved all of this out of Quickpatch and into its own component 👍

src/Game/Structs.hpp
```diff
@@ -11773,6 +11773,17 @@ namespace Game
 		unsigned int sysMsgTime;
 	};
 
+  struct AntilagClientStore
+	{
+    float realClientAngles[18][3];
+    float realClientPositions[18][3];
+    bool clientMoved[18];
```
Rackover, line 11780: MAX_CLIENTS

Rackover (changes requested):
This is just a preliminary review, I need to check every hook address and we'll probably have a bit of back and forth for the stubs to be rewritten (as diamante explained), but I like where this is going 😃

src/Game/Structs.hpp
L3D451R7: MAX_CLIENTS is defined in Functions.hpp which does not allow to reference it in Structs.hpp unfortunately.
Should I move it in Structs.hpp?

Rackover: I think so yes! It should be globally accessible

src/Components/Modules/AntiLag.cpp
L3D451R7: alright i went other route, had some problems with making __declspec(naked) functions so i went over inline asm in the Bullet_Fire func.

src/Components/Modules/AntiLag.cpp
L3D451R7: alright i went other path, tell me if there is any problem remaining

src/Components/Modules/AntiLag.cpp
```diff
@@ -0,0 +1,215 @@
...
+		// game does not run antilag on throwables except throwing knife.
+		if (!isThrowingKnife) {
+			return Game::G_RunMissileInternal(missile);
+		}
+
+		Game::AntilagClientStore antilagStore;
+
+		// i have no idea why it is dereferencing attachModelNames - thats what its actually pointing in this structure...
+		// the structure itself can be wrong. throwingDelay equals negative numbers like -67
+		int32_t throwingDelay = *reinterpret_cast<int32_t*>(&missile->attachModelNames[10]);
```
Caball009, line 186: I think this is because the current gentity struct layout is incorrect. I updated it here: https://github.com/Caball009/iw4x-client/commit/a22524bafe3eaa7b5dbf4caa87711d53b77814b4

The offset 0x1CC, as used in G_RunMissile, should now correctly point to missile->entData.missile.antilagTimeOffset.

L3D451R7: You're right, I can totally confirm with your changes its pointing to antilagTimeOffset which is accurate in this context. Thank you.

Rackover (changes requested):
In addition to the comments, while testing I got a hang while rotating to `devmap mp_rust`  after playing on afghan with bots. I could not reproduce it, although it's a bit concerning it might be my set up. I'll keep an eye open in case it happens again (i'll attach a debugger next time)

src/Components/Modules/Bullet.cpp
```diff
@@ -121,9 +206,15 @@ namespace Components
 			0.0f, std::numeric_limits<float>::max(), Game::DVAR_CODINFO, "Max range used when calculating the bullet end position");
 
 		Utils::Hook(0x4F6980, BG_GetSurfacePenetrationDepthStub, HOOK_JUMP).install()->quick();
-		Utils::Hook(0x440340, Bullet_FireStub, HOOK_JUMP).install()->quick();
 
-		Utils::Hook(0x440368, BG_srand_Hk, HOOK_CALL).install()->quick();
+		// we dont need this anymore since we rebuild Bullet_Fire
+		//Utils::Hook(0x440368, BG_srand_Hk, HOOK_CALL).install()->quick();
+
+		Utils::Hook(0x4A4E6B, Bullet_Fire, HOOK_CALL).install()->quick(); // FireWeapon
```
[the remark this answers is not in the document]
Laupetin: Without having a lot of context on whether rewriting a function here would make sense or not, there are a couple of advantages of rewriting a function instead of hooking it in the middle of its execution that might be worth considering:

* Hooking in the middle of a function is a lot more complex and increases the risk of making a mistake. I can remember fixing a bunch of these "random crashes" issues a few years back for example (patches from 4d1) that did exactly that, have been in some clients for ages and just had a minor mistake in the asm that most of the time did not lead to a crash but just then sometimes did on a map rotate for example. With things like jumps and tests you might destroy assertions that the compiler did about the state of flags, even when you perfectly recreate the registers state. Compared to that, the ABI between function calls is often a lot more lenient and easier to understand.

* Recreating a function gives you the context in which your modification is called. Reviewing code that is called within a function requires you to more closely and more often lookup the iw4 binary and is therefore harder and takes more time. This leads to sometimes patches not being examined anymore in the future (do we still need this? does this still do, what it is supposed to?) and avoided to be refactored.

* More complicated modifications (like moderate modifications to the execution flow) or multiple patches for multiple usecases often require multiple hooks inside the same function, increasing the effort of making a modification significantly compared to modifying a reimplementation.

My personal (current) way to go is to prefer reimplementations and only hooking at function call level unless it is very unfeasible because a function is too long, too complex or the modification is very minor so a simple patch is sufficient.

But that doesn't mean it has to be the approach for this project. I just wanted to give a few reasons to consider because i feel like generalizing a rule like 
> whenever modifying the game, the minimal amount of code should be rewritten and only when it is absolutely necessary

might not be consider this. In the end it is a trade-off of course though because your point are valid as well (especially the one with accidentally disabling other patches).

Rackover:
(Additionally - i suppose g_antilag should be an enum rather than an int?)

src/Components/Modules/Bullet.cpp
```diff
@@ -96,8 +84,17 @@ namespace Components
...
+		float spread,
+		Game::weaponParms* wpParms,
+		Game::gentity_s* weaponEnt,
+		Game::PlayerHandIndex handIndex,
+		int gameTime)
 	{
+		Game::AntilagClientStore antiLag_Backup;
+		AntiLag::G_AntiLagRewindClientPos(attacker, gameTime, &antiLag_Backup);
+
+#ifdef DEBUG_RIOT_SHIELD
```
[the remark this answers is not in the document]
L3D451R7: Left it as it was before my changes, if it needs to be toggleable let me know

src/Components/Modules/Bullet.cpp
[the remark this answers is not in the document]
L3D451R7: I went with rebuilding the thing entirely so I can debug it properly (seeing all the input and output for example), and I was taught to avoid Assembly entirely in code - it's not a C++ practice (I dont mean that everyone should follow the things i was taught - just explaining why I went with solution you can see). 

Since the original code in PR was using return address and stack hacks - I added assembler code to avoid those hacky things.

I will look into rumble issue soon, 4000 km away from home right now :)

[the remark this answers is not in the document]
L3D451R7: Also the call conversions are nasty in some functions, rebuilding was simpler than adding dirty inline assembly.

PR #171, opened 2024-12-03
feat: allow dedis to edit cheat dvars

### LINKED ISSUE #61: Uncheat player_sprintunlimited

diamante0018:
**What does this PR do?**

Fixes #61 

Dedicated servers will be able to edit cheat protected dvars

Draft, only testing this patch for now. It may have unintended consequences

### DIFF
--- a/src/Components/Modules/Dvar.cpp
+++ b/src/Components/Modules/Dvar.cpp
@@ -393,6 +393,49 @@ namespace Components
 		return dvar->domain.enumeration.strings[dvar->current.integer];
 	}
 
+	bool Dvar::CanEditCheatProtectedDvar(int flags)
+	{
+		// Allow the host to override cheat protected dvars
+		if (((flags & Game::DVAR_CHEAT) == Game::DVAR_CHEAT) && (*Game::sv_running && (*Game::sv_running)->current.enabled))
+		{
+			return true;
+		}
+
+		return false;
+	}
+
+	__declspec(naked) void Dvar::Dvar_IsCheatProtected_Stub() // Guessed name
+	{
+		__asm
+		{
+			push eax
+			pushad
+
+			push ecx // flags
+			call CanEditCheatProtectedDvar
+			add esp, 0x4
+
+			mov [esp + 0x20], eax
+			popad
+			pop eax
+
+			test al, al
+			jnz can_edit
+
+			// Call the original function and leave
+			push 0x64767D // Return address
+			push 0x646510 // Dvar_IsCheatProtected_Stub
+			ret
+
+		can_edit:
+			// When Dvar_IsCheatProtected_Stub returns 0 the game will allow us to change the dvar
+			xor eax, eax
+
+			push 0x64767D // Return address
+			ret
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


diamante0018:
@Rackover Hi, I completed testing this PR pre and post patch and I concluded there are no obvious behaviour changes to the client experience when it comes to dvar handling and replication for cheat protected dvars after the patch was applied to the dedicated server.
This is also the smallest patch we can make without hooking the dvar_setvariant function and perform the check there, however i dont think we need to do that because i found no changes in behaviour when done this way, in fact i guess this patch, is more efficient because it is applied once on dvar register rather than check every single time a dvar is getting updated if we are a dedi or not. so i guess in the end this is more efficient

Rackover:
This looks very elegant indeed! 
I think I like it, but would it be okay to make it not specific to dedicated servers? Listen servers should enjoy this.
I suppose you mean to apply this to the "host" in general, not just to dedis.

If you can make this check Game::sv_running instead of Dedicated::IsEnabled, then that should be good to go

Rackover:
Nevermind, this does not look related to #61 ?

https://github.com/iw4x/iw4x-client/issues/61#issuecomment-1818019268

This is the feature in question.  Uncheating every DVar on load is not acceptable, it does not even cover use cases for players where cheats are enabled or when a server owner wants to toggle a cheat.

Joining my own dedi server with this DLL does not work as intended. ❌

image

diamante0018:
I didn't really know the scope of this feature request was to toggle noclip commands, I thought the scope was simply just an adjustment on how the host of a game can set cheat protected commands without enabling sv_cheats (then every player can noclip not just the host)

diamante0018:
Anyhow, I believe the requets is now completed with the last commit. It simply does what the issue states which is to allow host and server to change cheat dvars without requiring sv_cheats to be set to 1.

diamante0018:
So I have studied how this behaviour was implemented on other clients where sv_cheats was missing because it was stripped on ship builds.

So there, in order even for the host to edit any cheat protected dvars sv_cheats must be 1

So that means that the moment the host can freely change sv_cheats to 1 the client can do cheat-protected commands such as any dvar change and then use god and noclip commands, why is this undesirable and not how this patch behaves?
Because the host if forced to allow too much to the client before they can edit anything.

My patch allows the host to edit any CHEAT dvars only but does not allow for any other cheat-protected command to be executed.

wroyca:
Converted to draft - need more discussion with louve (e.g. https://discord.com/channels/1110531063161299074/1200082418481250374/1332409817725272158)

diamante0018:
I would like to withdraw from this PR as well, I, unfortunately, did not understand the requirements set out in the issue and made a PR that does not achieve those requirements. On Discord there is a more detailed discussion which I will copy and paste into the issue so it's more accessible and anyone else pursuing that feat can do those things. thanks.
Time is scarce these days.

PR #173, opened 2024-12-05
Demo playback - stuck ammo counters workaround

Caball009:
**What does this PR do?**

During demo playback the hud isn't updated when changing weapons, so the ammo (clip) counters seem stuck.

This code is only actively doing something during demo playback, and should only provide a workaround for the above issue and nothing else.

If it's desirable to not show the weapon name upon weapon change (and only the ammo counters), then the cg->weaponSelectTime field shouldn't be set.


Caball009:
I'm not sure how much sense it makes to apply this to dedicated servers as well, but that goes for most stuff in the Theatre component.

src/Components/Modules/Theatre.cpp
```diff
@@ -426,5 +426,23 @@ namespace Components
...
+
+			if (auto* cg = Game::cgArray; !std::cmp_equal(cg->weaponSelect, cg->predictedPlayerState.weapCommon.weapon))
+			{
+				// change ammo counter and ammo clip counter
+				cg->weaponSelect = cg->predictedPlayerState.weapCommon.weapon;
+
+				// show weapon name
+				cg->weaponSelectTime = cg->time;
+			}
+		}, Scheduler::Pipeline::MAIN);
```
L3D451R7, line 446: shouldn't we just hook 3 places (names arent correct, just assumption: 0x449DE0 - SwitchToNextWeapon, 0x462540 - SwitchToPrevWeapon, 0x48BAF0 - BG_SelectWeapon, and maybe 0x472FF0 - should be ResetPlayerState) to avoid calling this code that much often?

L3D451R7, line 446: I think we can avoid this with removal of some demoType checks in functions that reference 0x59ED40.

L3D451R7: image

image

L3D451R7: I researched this a bit more, EV_WEAPON_SWITCH_STARTED gets passed in 0x4DD219 when viewing a demo, but its doing nothing and returning from this function, probably we could do this code there..

L3D451R7: image

diamante0018: I agree that a scheduler call should be avoided if this patch can be done in a hook when it's the right time in the engine to perform a fix

Caball009: I'd prefer a clean scheduled task to a hook that requires a stub with inline assembly, and maybe even an extra function address. It could be done in existing hook maybe, but I'm not sure if I like the idea of performing this specific task in an unrelated hook.

As for correctness, I don't think it matters much where this code is executed. The weapon selection fields won't be overwritten because they aren't written to in the first place, except when respawning or picking up weapons.

Rackover:
https://github.com/iw4x/iw4x-client/pull/173#discussion_r1873178916

Sorry but I agree with @L3D451R7  and @diamante0018  here - rather than a Scheduled task on the main thread I'd rather have this code called at the right time from the right place, systematically, by using a small hook on weapon change or on event reception -depending which is relevant.

I have not researched this issue a whole lot but if it needs be refreshed "on weapon change" then there's likely an event you can hook into to execute this code. I have no comments otherwise, the code looks good, but Scheduler is not the way to go for something that is bound to happen "in conjunction with something else" (here weapon switch). It's suited for stuff that is meant to happen every frame, which is probably a bit extra here.  I hope that makes sense

Rackover:
Ring me for another review when you've had time to address that 🙂   No emergency

PR #175, opened 2024-12-10
fix: restore lobby server / wider crashes

### LINKED ISSUE #174: Lobby Server unusable on the latest
Use a Lobby Server

The dedicated mode is no longer usable due to null-pointer check in-game code causing a crash

**What have you tried so far? (and my personal comments which are suggestions)**
This change needs to be reverted. Only put string dvar in these lambda functions. Do not change the old dvar handling code because you never know how this will affect large projects like this one.
image

Latest

Add Lobby servers to QA sessions

diamante0018:
**What does this PR do?**

Fixes #174 

~~Removes a command that executes a cfg that was entirely fabricated by the minds of the OG iw4x's devs and does not exist at all~~ (Reverted: see below) 
image (1)

The dvar mess is now gone, the private match dvar is registered by the live_init function and our custom privateserver function is registered by us because we use it in the info handlers

### DIFF
--- a/src/Components/Modules/Dedicated.cpp
+++ b/src/Components/Modules/Dedicated.cpp
@@ -62,7 +62,7 @@ namespace Components
 
 	void Dedicated::PostInitialization()
 	{
-		Command::Execute("exec autoexec.cfg");
+		Command::Execute("exec autoexec.cfg"); // Can be used by mods / server owners at will : Currently shows an error message on a default setup
 		Command::Execute("onlinegame 1");
 		Command::Execute("exec default_xboxlive.cfg");
 		Command::Execute("xblive_rankedmatch 1");
--- a/src/Components/Modules/Party.cpp
+++ b/src/Components/Modules/Party.cpp
@@ -213,9 +213,14 @@ namespace Components
 		Events::OnDvarInit([]
 		{
 			ServerVersion = Dvar::Register<const char*>("sv_version", "", Game::DVAR_SERVERINFO | Game::DVAR_INIT, "Server version");
-			PartyEnable = Dvar::Register<bool>("party_enable", Dedicated::IsEnabled(), Game::DVAR_NONE, "Enable party system");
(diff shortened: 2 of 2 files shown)


diamante0018:
Lobby server is fixed

image

diamante0018:
May I ask in which context there are used?

mxve:
Usually its meant for setting custom values to variables that aren’t available through menu options.
I think the idea is to keep the default config file unmodified and using the autoexec to overwrite its values. The autoexec wouldn’t be touched by a game when saving settings, so nothing custom would be overwritten. Also they are probably easier to work with as they only contain the modified values.

diamante0018:
Well, perhaps I did rush that change without thinking too much about it, while I do think my questions are valid I should probably revert that change just because this might break a mod from 1969 that may use this odd feature

I do believe these falls under the category of patches that no one knows about, undocumented, 99% of the servers owners do not know about this and I bet mod makers do not either, i think no one knew about this except at one point someone figured this feature existed and decided to use it. Well I guess i will add it back and leave someone else to think about it.

diamante0018:
This issue is wider than Lobby Servers. It also crashes sporadically when opening the private match menu and with mods.
All have the same crash address

diamante0018:
To recap what this code does:
Fixes a big nullptr crash that disables lobby servers and can trigger else where on some condition during normal game play (private match)

Cleans up the code around the dvars in question as a means to have cleaner and commented patches in party.cpp

Rackover (approved):

Rackover:
Tested on dedi, private match, lobby server - works.
Tested online and produces no regression as far as i can see

Rackover:
Well, looks like I missed a regression.
image
image-1

You're free to re-submit your fix ontop of the revert in a new PR, this time I will pay more attention.
Sorry for the back-and-forth, regressions like this are easy to miss on initial testing if I don't pay too much attention to the console.

PR #180, opened 2024-12-15
Fix gamepad performance

wroyca:
Fix performance issue from gamepad under specific scenario (e.g. running under wine). 

Note: under 10 fps will cause trigger roles to lock but this should be considered as a different issue.

### DIFF
--- a/src/Components/Modules/Gamepad.cpp
+++ b/src/Components/Modules/Gamepad.cpp
@@ -2009,7 +2009,7 @@ namespace Components
 		api.SetForceFeedback(triggerFeedbacks[LEFT], triggerFeedbacks[RIGHT]);
 	}
 
-	Gamepad::Gamepad()
+	Gamepad::Gamepad() : run(true)
 	{
 		if (ZoneBuilder::IsEnabled())
 		{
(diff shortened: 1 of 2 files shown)


src/Components/Modules/Gamepad.cpp
```diff
@@ -2079,26 +2079,38 @@ namespace Components
...
+						std::this_thread::sleep_for(remainingTime);
+					else
+						std::this_thread::yield();
+				}
+			});
 		}
 	}
+
+	Gamepad::~Gamepad()
+	{
```
Caball009, line 2110: This can be achieved in a cleaner way fwiw: `std::jthread` + `std::stop_token` / `std::stop_source`.
https://en.cppreference.com/w/cpp/thread/jthread/get_stop_source

wroyca: On paper yes; but introducing std::jthread would require me to invest additional time and effort to fully understand its nuances, integrate it properly, AND ensure it doesn’t introduce unexpected side effects/compatibility issues. For now, addressing the immediate problem with the existing std::thread implementation is the most practical approach (IMO). ;)

Caball009: Fair enough.

src/Components/Modules/Gamepad.hpp
```diff
@@ -24,6 +30,7 @@ namespace Components
 		static const int RUMBLE_CONFIGSTRINGS_COUNT = 32;
 
 		Gamepad();
+	 ~Gamepad();
```
diamante0018, line 33: Something afoot here happened with formatting. Also I noticed that maybe the inlcudes could go to stdinclude.hpp but it's not a big deal

diamante0018 (commented):
good pr

src/Components/Modules/Gamepad.hpp
wroyca: I did it on purpose for alignment but forgot to change it back before committing. Fixed, thanks!

Rackover (approved):

Rackover:
Tested and runs as intended, no performance impact - on Win10 x64

PR #181, opened 2024-12-17
[Stats]: Fix code and save stats to mods folder when needed

diamante0018:
**What does this PR do?**

This is a big fix, as the title says. The code is restored to how it was in 2022 prior to one unintentional change: strncmp was accidentally replaced with strcmp, causing the code never to recognize a mod is present.

Restore intended behaviour

No


diamante0018:
good 
image

src/Components/Modules/Stats.cpp
```diff
@@ -69,9 +69,12 @@ namespace Components
 	{
 		assert(*Game::fs_gameDirVar);
 
-		if (!std::strcmp((*Game::fs_gameDirVar)->current.string, "mods/"))
+		static std::string modsFolder; // This var is static to ensure the string pointer does not become nullptr after this function exits
```
wroyca, line 72: It's ok to drop static here, the real issue (if any) could arise from using the pointer (c_str()) outside the function while the string itself is modified, but that's unrelated to the static keyword :saluting_face:

wroyca (changes requested):

src/Components/Modules/Stats.cpp
diamante0018: Sure. I'll apply the change tomorrow

diamante0018: I  retraced my steps a bit and pushed a new commit, I think this should be good because there is no static  var and fixes the issue

diamante0018:
image
confirm it's working

wroyca (approved):

Issue #177, opened 2024-12-13
Raw input sends the mouse flickering to outer space


Rackover:
Might be related to https://github.com/iw4x/iw4x-client/pull/166 
Or https://github.com/iw4x/iw4x-client/pull/148 ?

Have not investigated yet. Feature is disabled for now as a temporary measure

Rackover:
https://github.com/iw4x/iw4x-client/commit/92f543682a8dc8adf0d042c1258758aa3718c6a2

Rackover:
https://github.com/user-attachments/assets/778a4f9b-bec7-4147-9577-1592f9d94795

wroyca:
cc @L3D451R7, I noticed that your changes accidentally broke cursor clipping (ClipCursor)—potentially related?

L3D451R7:
> cc @L3D451R7, I noticed that your changes accidentally broke cursor clipping (ClipCursor)—potentially related?

I have caught the issue myself after I did repeat everything from video above, looking into it - i'll try to get it fixed on this week.

L3D451R7:
@Rackover Does it continue to happen if reverting #148 ? I had not alot of time to investigate but that was first what should be tested since game doesn't do this. I tried myself and couldn't reproduce the issue but it's might be just me having problems with reproducting it.

Rackover:
If these two features interact in a way that destroys either we need to know why to know how to de-entangle them. I will try with 148 reverted but even if it works then, it still won't be enough to backtrack 148 in favour of this one - we'll need to work out a way for them to coexist, or understand exactly why they won't

wroyca:
@L3D451R7 @Rackover 

In RawMouse::IN_Frame(), GetForegroundWindow is incorrectly used to determine focus. GetForegroundWindow only identifies which window is in the foreground (under the cursor), not whether the application is focused. When the camera is moved left or right in-game (in windowed mode), the cursor moves transparently, which is expected, and more importantly, it will end up moving outside the window. ClipCursor then corrects the cursor's position back into the window. That brief timeframe before the correction means that GetForegroundWindow will return false, which in turn causes ToggleRawInput to be turned off. Likewise, once ClipCursor makes its correction, GetForegroundWindow becomes true again, triggering ToggleRawInput to be turned on. Forcing ToggleRawInput(true) in IN_Frame() will prevents the toggle dance, but breaks window controls (min, max, close), among others.

[...]

PR #186, opened 2024-12-23
Raw Input Flickering Fix | Issue #177

L3D451R7:
reworked IN_RecenterMouse, improved focus checks for raw input, fixed Raw Input bugs caused by mouse reaching edges of application window rect (#177)

### DIFF
--- a/src/Components/Modules/RawMouse.cpp
+++ b/src/Components/Modules/RawMouse.cpp
@@ -33,6 +33,7 @@ namespace Components
 	uint32_t RawMouse::MouseRawEvents = 0;
 
 	bool RawMouse::InRawInput = false;
+	bool RawMouse::InFocus = false;
 	bool RawMouse::FirstRawInputUpdate = true;
 
 	constexpr const int K_MWHEELUP = 205;
@@ -164,7 +165,7 @@ namespace Components
 		if (!GetRawInput(lParam, raw, dwSize))
 			return TRUE;
 
-		if (GetForegroundWindow() != Window::GetWindow())
+		if (!RawMouse::InFocus)
 			return TRUE;
 
 		// Is there's really absolute mouse on earth?
@@ -231,14 +232,13 @@ namespace Components
 
 		if (M_RawInput.get<bool>())
 		{
-			if (MouseEvents == 0)
-				return TRUE;
+			if (MouseEvents == 0 || !RawMouse::InFocus)
+				return FALSE;
 
 			if (M_RawInputVerbose.get<bool>())
 				Logger::Debug("Window Mouse Message: [{}, {}]", MouseEvents, MouseRawEvents);
 
 			MouseRawEvents = MouseEvents;
-			return TRUE;
 		}
 
 		Game::IN_MouseEvent(MouseEvents);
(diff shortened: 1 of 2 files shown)


src/Components/Modules/RawMouse.cpp
```diff
@@ -369,27 +378,30 @@ namespace Components
...
-		ToggleRawInput(focused);
+		if (RawMouse::InFocus)
+			ToggleRawInput(IsMouseInClientBounds());
+
 		return Game::IN_Frame();
 	}
 
 	BOOL RawMouse::IN_RecenterMouse()
 	{
-		RECT clientRect;
```
wroyca, line 381: This is not correct. ClipCursor expects the rectangle's coordinates to be in screen coordinates, which means the coordinates must reflect their position on the screen rather than within the window's client area, hence ClientToScreen

wroyca (changes requested):

src/Components/Modules/RawMouse.cpp
```diff
@@ -287,13 +289,20 @@ namespace Components
...
-
-		ClipCursor(NULL);
+		if (!Game::CL_MouseEvent(curPos.x, curPos.y, dx, dy))
+		{
+			ClipCursor(NULL);
+			return;
+		}
 
-		if (recenterMouse)
+		RECT Rect;
```
wroyca, line 298: https://github.com/iw4x/iw4x-client/pull/186#discussion_r1896883907

src/Components/Modules/RawMouse.cpp
```diff
@@ -424,13 +436,14 @@ namespace Components
 
 				prevCursorPos.x = WindowCenterX;
 				prevCursorPos.y = WindowCenterY;
+
+				ClipCursor(&Rect);
```
wroyca, line 440: https://github.com/iw4x/iw4x-client/pull/186#discussion_r1896883907

src/Components/Modules/RawMouse.cpp
L3D451R7: I figured out the function is missing the important code which is doing the mouse recenter, it's only performing clipping which is actually might be a real reason why the flicking is happening in first place.

src/Components/Modules/RawMouse.cpp
```diff
@@ -287,11 +289,14 @@ namespace Components
 
 		Gamepad::OnMouseMove(curPos.x, curPos.y, dx, dy);
 
-		auto recenterMouse = Game::CL_MouseEvent(curPos.x, curPos.y, dx, dy);
-
-		ClipCursor(NULL);
+		if (!Game::CL_MouseEvent(curPos.x, curPos.y, dx, dy))
```
wroyca, line 292: Why check CL_MouseEvent to disable clipping? ClipCursor(NULL) was originally used to prevent clipping in the menu. Does CL_MouseEvent return false in menus and true otherwise?

If so, we should add a comment to clarify. :)

wroyca (changes requested):

src/Components/Modules/RawMouse.cpp
L3D451R7: It returns true when it wants mouse to be recentered.

Rackover:
I see no obvious issue with the code, the get<dvar_t*> thingies are a bit odd but I don't mind them.
Testing it, I cannot reproduce the bug, and the gamepad seems to work gracefully. although, I cannot perceive a difference between `m_rawInput 1` and `m_rawInput 0` anymore (it used to feel really different, now it doesn't)

Confirm to me that everything is working as intended?
If so, I'm ready to merge this as I see no glaring undesirable side effect compared to master, so I'm ready to trust you it's a change for the better 🙂

Rackover (approved):

wroyca:
Hello! Here are some issues I noticed during my test:

The cursor takes too long to clip back when focus is lost and regained. This delay often cause the cursor to escape the window accidentally while aiming:

https://github.com/user-attachments/assets/21a15eb1-27af-42cb-9224-d9b681270c23

The cursor moves in the background, meaning its position is not fixed. It appears in different positions when switching back and forth between the menu and the game. Occasionally, the cursor position is recalculated, leading to erratic jumps within the menu:

https://github.com/user-attachments/assets/bc5b4b1f-b37c-4e4a-9273-79d5dc86a07c

PR #188, opened 2025-01-03
fix logfile creation with multiple instances

### LINKED ISSUE #187: IW4x cannot create more than one log file (when I-th instances are opened)
IW4(x) is unable to create more than one file meaning when two instances of the game run the game will fail to open a log file and/or will write to the same logfile because the return value of FS_FRotate is not checked (on IW4 it does not even seem to return a value if that means anything)

IW6 fixed this by giving the game 16 more chances to find an available file log to rotate and write to.

Indicative that IW realized this was an issue when they started hosting game servers themselves and they had more than one dedi running on the same machine.

...

...

Engine issue

A PR will soon come

diamante0018:
**What does this PR do?**

Fixes #187

For the best

### DIFF
--- a/src/Components/Modules/FileSystem.cpp
+++ b/src/Components/Modules/FileSystem.cpp
@@ -378,6 +378,65 @@ namespace Components
 		return Utils::Hook::Call<Game::HunkUser*(int, const char*, bool, int)>(0x430E90)(maxSize, name, fixed, type);
 	}
 
+	bool FileSystem::FileWrapper_Rotate(const char* ospath)
+	{
+		constexpr auto MAX_BACKUPS = 20;
+
+		std::string renamedPath;
+
+		std::optional<int> oldestIndex;
+		auto currentIndex = 0;
+		std::filesystem::file_time_type oldestime{};
+
+		// Check if the original file exists
+		if (!Utils::IO::FileExists(ospath))
+		{
+			return true; // Return true if the file does not exist (no file to rotate)
+		}
+
+		for (; currentIndex < MAX_BACKUPS; ++currentIndex)
+		{
+			renamedPath = std::format("{0}.{1:03}", ospath, currentIndex);
+
+			if (!Utils::IO::FileExists(renamedPath))
+			{
+				break; // Stop if an available slot is found
+			}
+
+			auto time = std::filesystem::last_write_time(renamedPath);
+			if (!oldestIndex.has_value() || time < oldestime)
+			{
+				oldestime = time;
+				oldestIndex = currentIndex;
+			}
+		}
+
+		if (currentIndex == MAX_BACKUPS)
+		{
+			renamedPath = std::format("{0}.{1:03}", ospath, *oldestIndex);
+			Utils::IO::RemoveFile(renamedPath); // Remove the oldest backup file
+		}
+		else
+		{
+			renamedPath = std::format("{0}.{1:03}", ospath, currentIndex);
+		}
+
+		// Rename the original file to the selected backup slot
+		std::error_code ec;
+		std::filesystem::rename(ospath, renamedPath, ec);
+
+		return !ec;
+	}
+
+	bool FileSystem::FileRotate(const std::string& filename)
+	{
+		std::array<char, MAX_OSPATH> ospath{};
+
+		const auto* basepath = (*Game::fs_homepath)->current.string;
+		Game::FS_BuildOSPath(basepath, Game::fs_gamedir, filename.c_str(), ospath.data());
+		return FileWrapper_Rotate(ospath.data());
+	}
+
 	FileSystem::FileSystem()
 	{
 		// Thread safe file system interaction
--- a/src/Components/Modules/Logger.cpp
+++ b/src/Components/Modules/Logger.cpp
@@ -1,4 +1,5 @@
 
+#include "Branding.hpp"
 #include "Console.hpp"
 #include "Events.hpp"
 
@@ -37,9 +38,9 @@ namespace Components
 	void Logger::MessagePrint(const int channel, const std::string& msg)
 	{
 		static const auto shouldPrint = []() -> bool
-			{
-				return Flags::HasFlag("stdout");
-			}();
+		{
+			return Flags::HasFlag("stdout");
+		}();
 
 			if (shouldPrint)
 			{
@@ -119,33 +120,33 @@ namespace Components
 	void Logger::PrintFail2BanInternal(const std::string_view& fmt, std::format_args&& args)
 	{
 		static const auto shouldPrint = []() -> bool
-			{
-				return Flags::HasFlag("fail2ban");
-			}();
+		{
+			return Flags::HasFlag("fail2ban");
+		}();
 
-			if (!shouldPrint)
-			{
-				return;
-			}
+		if (!shouldPrint)
+		{
+			return;
+		}
 
-			auto msg = std::vformat(fmt, args);
+		auto msg = std::vformat(fmt, args);
 
-			static auto log_next_time_stamp = true;
-			if (log_next_time_stamp)
-			{
-				auto now = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
-				// Convert to local time
-				std::tm timeInfo = *std::localtime(&now);
+		static auto log_next_time_stamp = true;
+		if (log_next_time_stamp)
+		{
+			auto now = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+			// Convert to local time
+			std::tm timeInfo = *std::localtime(&now);
 
-				std::ostringstream ss;
-				ss << std::put_time(&timeInfo, "%Y-%m-%d %H:%M:%S ");
+			std::ostringstream ss;
+			ss << std::put_time(&timeInfo, "%Y-%m-%d %H:%M:%S ");
 
-				msg.insert(0, ss.str());
-			}
+			msg.insert(0, ss.str());
+		}
 
-			log_next_time_stamp = (msg.find('\n') != std::string::npos);
+		log_next_time_stamp = (msg.find('\n') != std::string::npos);
 
-			Utils::IO::WriteFile(IW4x_fail2ban_location.get<std::string>(), msg, true);
+		Utils::IO::WriteFile(IW4x_fail2ban_location.get<std::string>(), msg, true);
 	}
 
 	void Logger::Frame()
@@ -233,21 +234,21 @@ namespace Components
 			popad
 			ret
 
-			returnPrint :
+		returnPrint:
 			pushad
 
-				push 0 // gLog
-				push[esp + 2Ch] // data
-				call NetworkLog
-				add esp, 8h
+			push 0 // gLog
+			push[esp + 2Ch] // data
+			call NetworkLog
+			add esp, 8h
 
-				popad
+			popad
 
-				push esi
-				mov esi, [esp + 0Ch]
+			push esi
+			mov esi, [esp + 0Ch]
 
-				push 4AA835h
-				ret
+			push 4AA835h
+			ret
 		}
 	}
 
@@ -438,6 +439,44 @@ namespace Components
 		}
 	}
 
+	void Logger::Com_OpenLogFile_Stub()
+	{
+		// 16 should be enough as realistically someone will probably not run as many servers
+		// the code below will try 16 times to find an open slot between any of the 'backup' slots
+		constexpr auto MAX_LOGFILE_CREATE_ATTEMPTS = 16;
+
+		std::string logfile;
+
+		if (!Game::Sys_IsMainThread() || *Game::opening_qconsole)
+		{
+			return;
+		}
+
+		*Game::opening_qconsole = true;
+
+		for (auto i = 0; i < MAX_LOGFILE_CREATE_ATTEMPTS; ++i)
+		{
+			logfile = (i == 0) ? Game::logFileName : std::format("{0}.{1:03}", Game::logFileName, i);
+
+			if (!FileSystem::FileRotate(logfile))
+			{
+				continue;
+			}
+
+			*Game::logfile = Game::FS_FOpenTextFileWrite(logfile.c_str());
+			if (*Game::logfile)
+			{
+				Game::Com_Printf(Game::CON_CHANNEL_SYSTEM, "\'%s\'\n", Game::Com_GetCommandLine());
+				const auto time = Utils::GetTime();
+				Game::Com_Printf(Game::CON_CHANNEL_SYSTEM, "Build %s. Logfile opened on %s\n", Branding::GetBuildNumber(), time.c_str());
+				break; // Stop attempting further backups
+			}
+		}
+
+		*Game::opening_qconsole = false;
+		*Game::com_consoleLogOpenFailed = *Game::logfile == 0;
+	}
+
 	Logger::Logger()
 	{
 		// Print sound aliases errors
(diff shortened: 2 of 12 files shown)


src/Components/Modules/FileSystem.cpp
```diff
@@ -379,6 +379,62 @@ namespace Components
...
+
+		// Check if the original file exists
+		if (_stat(ospath, &statBuf))
+		{
+			return true; // Return true if the file does not exist (no file to rotate)
+		}
+
+		for (; currentIndex < MAX_BACKUPS; ++currentIndex)
+		{
+			(void)sprintf_s(renamedPath, "%s.%03i", ospath, currentIndex);
```
wroyca, line 401: nit: don't format, could be a variable, e.g. pass currentIndex with https://en.cppreference.com/w/cpp/string/basic_string/to_string

src/Components/Modules/FileSystem.cpp
```diff
@@ -379,6 +379,62 @@ namespace Components
...
+
+			if (_stat(renamedPath, &statBuf))
```
wroyca, line 403: nit: https://en.cppreference.com/w/cpp/filesystem/exists

src/Components/Modules/FileSystem.cpp
```diff
@@ -379,6 +379,62 @@ namespace Components
...
+			(void)sprintf_s(renamedPath, "%s.%03i", ospath, oldestIndex);
+			(void)std::remove(renamedPath); // Remove the oldest backup file
+		}
+		else
+		{
+			(void)sprintf_s(renamedPath, "%s.%03i", ospath, currentIndex);
+		}
+
+		// Rename the original file to the selected backup slot
+		return std::rename(ospath, renamedPath) == 0;
```
wroyca, line 426: nit: don't format, don't use st_mtime, prefer https://en.cppreference.com/w/cpp/filesystem/last_write_time, https://en.cppreference.com/w/cpp/chrono/time_point/time_since_epoch; the whole block can be simplified

src/Components/Modules/FileSystem.cpp
```diff
@@ -379,6 +379,62 @@ namespace Components
...
+			{
+				oldestTime = statBuf.st_mtime;
+				oldestIndex = currentIndex;
+			}
+		}
+
+		if (currentIndex == MAX_BACKUPS)
+		{
+			(void)sprintf_s(renamedPath, "%s.%03i", ospath, oldestIndex);
+			(void)std::remove(renamedPath); // Remove the oldest backup file
```
wroyca, line 418: nit: https://en.cppreference.com/w/cpp/filesystem/remove

src/Components/Modules/FileSystem.cpp
```diff
@@ -379,6 +379,62 @@ namespace Components
...
+			(void)sprintf_s(renamedPath, "%s.%03i", ospath, currentIndex);
+		}
+
+		// Rename the original file to the selected backup slot
+		return std::rename(ospath, renamedPath) == 0;
+	}
+
+	bool FileSystem::FileRotate(const char* filename)
+	{
+		char ospath[MAX_OSPATH]{};
```
wroyca, line 431: nit: https://en.cppreference.com/w/cpp/container/array

src/Components/Modules/FileSystem.cpp
```diff
@@ -379,6 +379,62 @@ namespace Components
...
+		else
+		{
+			(void)sprintf_s(renamedPath, "%s.%03i", ospath, currentIndex);
+		}
+
+		// Rename the original file to the selected backup slot
+		return std::rename(ospath, renamedPath) == 0;
+	}
+
+	bool FileSystem::FileRotate(const char* filename)
```
wroyca, line 429: nit: const std::string& filename

src/Components/Modules/Logger.cpp
```diff
@@ -439,6 +440,47 @@ namespace Components
...
+
+		*Game::opening_qconsole = true;
+
+		tm new_time{};
+		_time64(&aclock);
+		_localtime64_s(&new_time, &aclock);
+
+		for (auto i = 0; i < 16; ++i)
+		{
+			const auto* file_name = (i == 0) ? "console_mp.log" : Utils::String::VA("console_mp.log.%03i", i);
```
wroyca, line 462: nit: logfile could just be a string, we don't need the whole I_strncpyz dance

src/Components/Modules/Logger.cpp
```diff
@@ -439,6 +440,47 @@ namespace Components
...
+			if (!FileSystem::FileRotate(logfile))
+			{
+				continue;
+			}
+
+			*Game::logfile = Game::FS_FOpenTextFileWrite(logfile);
+			if (*Game::logfile)
+			{
+				Game::Com_Printf(Game::CON_CHANNEL_FILES, "\'%s\'\n", Game::Com_GetCommandLine());
+				(void)asctime_s(time, sizeof(time), &new_time);
```
wroyca, line 474: nit: https://en.cppreference.com/w/cpp/chrono/c/strftime

src/Components/Modules/Logger.cpp
```diff
@@ -439,6 +440,47 @@ namespace Components
...
+		if (!Game::Sys_IsMainThread() || *Game::opening_qconsole)
+		{
+			return;
+		}
+
+		*Game::opening_qconsole = true;
+
+		tm new_time{};
+		_time64(&aclock);
+		_localtime64_s(&new_time, &aclock);
```
wroyca, line 458: nit: localtime_s, it will evaluate to _localtime64_s

src/Components/Modules/Logger.cpp
```diff
@@ -439,6 +440,47 @@ namespace Components
...
+
+		if (!Game::Sys_IsMainThread() || *Game::opening_qconsole)
+		{
+			return;
+		}
+
+		*Game::opening_qconsole = true;
+
+		tm new_time{};
+		_time64(&aclock);
```
wroyca, line 457: nit: rename aclock to currentTime and use https://en.cppreference.com/w/cpp/chrono/c/time instead of _time64

src/Components/Modules/Logger.cpp
```diff
@@ -439,6 +440,47 @@ namespace Components
 		}
 	}
 
+	void Logger::Com_OpenLogFile_Stub()
+	{
+		time_t aclock;
+		char time[32]{};
+		char logfile[MAX_OSPATH];
```
wroyca, line 447: nit: logfile could just be a string, see below

src/Components/Modules/Logger.cpp
Rackover: Is there no other place in the IW4x codebase where we convert a time to a string?
If there is any, how was it done?

src/Components/Modules/FileSystem.cpp
```diff
@@ -379,6 +379,65 @@ namespace Components
...
+			if (oldestIndex == -1 || time < oldestime)
+			{
+				oldestime = time;
+				oldestIndex = currentIndex;
+			}
+		}
+
+		if (currentIndex == MAX_BACKUPS)
+		{
+			(void)sprintf_s(renamedPath, "%s.%03i", ospath, oldestIndex);
```
Rackover, line 417: I know you like casting non-returning functions to `(void)` for whatever reason, but do you intend to convert the whole IW4x project to that convention or not?

Cause if not, then you can stop doing it right there - I'm not adding a redundant extra six characters in front of 90% of function calls just to please your aesthetics 😅

src/Components/Modules/Logger.cpp
diamante0018: IW4x code? I don't know
This original code was taken by the game's code and _s were added since we are on the windows platform and a lot of function have a _safe variant
image

src/Components/Modules/Logger.cpp
diamante0018: I think this project can use microsoft specific code because it's made to be only compiled by the msvc compiler. It's not something bad, we use Microsoft-specific functions everywhere around the code base, from winapi calls to hook API, to format apis, this is the least 'microsoft' specific egregious case where it's really not a big deal imo

src/Components/Modules/Logger.cpp
Rackover: If you don't know, you can look!
You're adding this code to the IW4x project, not me - make sure it is compliant / consistent with **iw4x code** and if there is no prior example, then make sure your code is **C++** whenever possible instead of C.
If you don't like strftime, another way to do it would be std::put_time into stringstream or something. I'm sure there are plenty of other ways.

You do not need to emulate the code style of the game - it was coded in a different language, at a different time, and for different platforms.

diamante0018: I suppose there is an identical code like this in theatre.cpp but since it's not in the scope of this PR I'll ignore it and write a new c++ function in utils.cpp to get the time and hopefully it will dispel any concern for micosoft specific evil functions and lack of c++ code in this pr.

diamante0018: I hope this new format will be accepted as being identical in functionality can be used going forward as a helper functions

image

Rackover: I agree, I think adding a helper function for this is the way to go 🙂

src/Components/Modules/FileSystem.cpp
diamante0018: "Unnecessary" *overcomplication* for what is a simple char array[]. This is in no way an improvement in this specific case, there is no reason to go bonkers on C++. This is the *same* as Louvenarde asking me not to cast return values to (void). This feels like the same way! 0 code improvement but this is way more than 6 characters added at the start of the function, isn't it?!
However, I can't help but notice a thumbs-up from Louvenarde so here goes nothing, I guess some code is seen in some different light sometimes

diamante0018:
There is so much C++ code in this PR now, you won't even believe how much C++ you can fit in this code.

Here is the final result:

image

Video evidence of file rotation over multiple instances on the latest commit:

https://youtu.be/r1DzFBZP4m8

diamante0018:
Here is a longer video showcasing the rotation over older files when the entire 20 slots for backed-up logfiles are filled up

https://youtu.be/di2pKWL0ptw

You can test it yourself with this code easily to create 20 backups as I'm sure this is also gonna get tested before the PR is merged by someone else 😄 

Be careful tho, this is C code, it might harm your PC in some unforeseeable ways.

`c
#include <stdio.h>

int main(void) {
    char filename[256];
    FILE *file;

    for (int i = 0; i < 20; ++i) {
        snprintf(filename, sizeof(filename), "console_mp.log.%03d", i);

        file = fopen(filename, "w");
        if (file) {
            fprintf(file, "This is file %d\n", i);
            fclose(file);
            printf("Created file: %s\n", filename);
        } else {
            perror("Error creating file");
            return 1;
        }
    }

    printf("All files created successfully.\n");
    return 0;
}
```

File: src/Components/Modules/FileSystem.cpp (hunk 5 above)
Comment (Rackover) [reply src/Components/Modules/FileSystem.cpp]: I will elaborate on my 👍 : 
- I agree with "We want to use modern practices where applicable", because in general I prefer the use of std::string& over const char* for instance, or std::chrono over C-style time functions, and so on and so on. So this is the sentence I'm approving
- Regarding the `char osPath[MAX_OSPATH]` vs `std::array<char>` , i have no opinion. I prefer the former because I'm more used to it, but it's no strong argument - if std::array has other benefits, they outweight my preference on this matter. 

Thus being said I feel like it's no big deal. It's a `nit:`  after all and maybe not a very important one. We use C-style array declarations all over the code, and arguably, they _are_ a bit easier to read than `std::array<T, S> {};`. It's fine either way.

Comment (diamante0018) [reply src/Components/Modules/FileSystem.cpp]: Okay, I appreciate the clarification.

File: src/Components/Modules/Logger.cpp (hunk 10 above)
Comment (wroyca) [reply src/Components/Modules/Logger.cpp]: Yes, the codebase is old and crufty, but that’s all the more reason not to double down on it.

File: src/Components/Modules/FileSystem.cpp [hunk 13]
```diff
@@ -379,6 +379,65 @@ namespace Components
 		return Utils::Hook::Call<Game::HunkUser*(int, const char*, bool, int)>(0x430E90)(maxSize, name, fixed, type);
 	}
 
+	bool FileSystem::FileWrapper_Rotate(const char* ospath)
+	{
+		constexpr auto MAX_BACKUPS = 20;
+
+		std::string renamedPath;
+
+		auto oldestIndex = -1;
```
Rackover, line 388: You can use std::optional<> rather than a magic "-1" value here (e.g. std::optional<size_t>)

src/Components/Modules/Logger.cpp
```diff
@@ -439,6 +440,40 @@ namespace Components
...
+		std::string logfile;
+
+		if (!Game::Sys_IsMainThread() || *Game::opening_qconsole)
+		{
+			return;
+		}
+
+		*Game::opening_qconsole = true;
+
+		for (auto i = 0; i < 16; ++i)
```
Rackover, line 454: Why is it 20 there and 16 here?

src/Components/Modules/Logger.cpp
```diff
@@ -439,6 +440,40 @@ namespace Components
...
+		{
+			logfile = (i == 0) ? "console_mp.log" : std::format("console_mp.log.{0:03}", i);
```
Rackover, line 456: doesn't the log file already get renamed during the rotation & rotation check?
so according to the code in FileRotate you should always give `console_mp.log` no, and let FileRotate add a numbered suffix ?

Rackover, line 456: grab from `0x730130` (`console_mp.log`) please

Rackover (changes requested):
okay well it's a bit odd but i'll take it.  I have not tested it yet, I leave it in the hands of wroy

src/Components/Modules/Logger.cpp
diamante0018: That's two separate mechanisms in action. File rotates functions finds the oldest log file to delete if necessary. This for loop finds the first log file out of 16 slots that can be used and is not already in used by another instance which is the whole point of this pr

diamante0018: 16 is magic value choose by IW when devviving iw6 servers. They probably figured 16 was the max number of servers they were gonna run and leave 4 log files slots as backup in total if it was necessary. Let me know if 20 or 16 should be used

Rackover: wait, so what's the difference between second slot and first rotation of first slot ?
both are `console_mp.001` no ?

diamante0018: Yeah.. If I'm understanding your question (magic value 16 and 20 aside), you wanted to know why there is a for loop here that runs 16 times. So basically, this section of the code will try to find a log file that is not in use by any other instance.
The fs_rotate function will now return a bool in case it found a file that is not in use (or does not exist between 000 and 019) if it successfully rotates a log file iw4x will start to write to the first empty slot between 001 and 019. the for loop that runs 16 times is there to ensure that enough attempts are given to the client to allow fs_rotate to find a free log file.

diamante0018:
@Rackover I addressed your last review

diamante0018:
I finished testing it again with the c code provided. all is good.

diamante0018:
@wroyca maybe you could leave a review too if all the conditions have been satisfied.

wroyca (approved):

wroyca (approved):

diamante0018:
Thank you

PR #192, opened 2025-01-07
fix flags for  rumble dvars

diamante0018:
**What does this PR do?**

Big fix for saving dvars

- [ ] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [ ] Mention any [related issues](https://github.com/iw4x/iw4x-client/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [ ] Follow our [coding conventions](https://github.com/iw4x/iw4x-client/blob/master/CODESTYLE.md)
- [ ] Minimize the number of commits

### DIFF
--- a/src/Components/Modules/Rumble.cpp
+++ b/src/Components/Modules/Rumble.cpp
@@ -1029,8 +1029,8 @@ namespace Components
 
 	void Rumble::InitDvars()
 	{
-		cl_debug_rumbles = Dvar::Register<bool>("cl_debug_rumbles", false, Game::DVAR_SAVED, "Debug rumbles on the screen");
-		cl_rumbleScale = Dvar::Register<float>("cl_rumbleScale", 0.6f, 0.f, 1.f, Game::DVAR_SAVED, "Rumble multiplier for the controller");
+		cl_debug_rumbles = Dvar::Register<bool>("cl_debug_rumbles", false, Game::DVAR_NONE, "Debug rumbles on the screen");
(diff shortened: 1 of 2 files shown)


src/Components/Modules/Rumble.cpp
```diff
@@ -1030,8 +1030,8 @@ namespace Components
 
 	void Rumble::InitDvars()
 	{
-		cl_debug_rumbles = Dvar::Register<bool>("cl_debug_rumbles", false, Game::DVAR_SAVED, "Debug rumbles on the screen");
-		cl_rumbleScale = Dvar::Register<float>("cl_rumbleScale", 0.6f, 0.f, 1.f, Game::DVAR_SAVED, "Rumble multiplier for the controller");
+		cl_debug_rumbles = Dvar::Register<bool>("cl_debug_rumbles", false, Game::DVAR_NONE, "Debug rumbles on the screen");
```
wroyca, line 1033: I say let's keep this as DVAR_SAVED for now, at least until we can observe and document DVAR_SAVED behavior. ;)

wroyca (changes requested):

src/Components/Modules/Rumble.cpp
diamante0018: That's not a good idea, archive flags needs to be added to save to the config file which should be the intended behaviour for dvar like this one

wroyca: > That's not a good idea, archive flags needs to be added to save to the config file which should be the intended behaviour for dvar like this one

This is a debug flag, it shouldn't be exposed in user config file

diamante0018: I see I'll look again, I think I missed that
I log back in to my pc tomorrow

wroyca (approved):

PR #194, opened 2025-01-07
Fix HAVE_UNISTD_H definition with MSBuild under Wine

wroyca:
Wine leads Premake to incorrectly detect unistd.h as available because it exposes the host environment (e.g., /usr/include/) while emulating a Windows system. Since Wine designates the host system as Window, we can ensure that HAVE_UNISTD_H is only set when Premake is running natively (e.g., when generating MinGW Make config)

### DIFF
--- a/deps/premake/zlib.lua
+++ b/deps/premake/zlib.lua
@@ -16,7 +16,7 @@ function zlib.includes()
 		"ZLIB_CONST",
 	}
 
-	if os.isfile("/usr/include/unistd.h") then
+	if os.host() ~= "windows" and os.isfile("/usr/include/unistd.h") then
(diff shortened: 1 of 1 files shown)


Rackover:
Does this not change the behaviour on windows, instead of just on wine?

wroyca:
> Does this not change the behaviour on windows, instead of just on wine?

It doesn't since /usr/include/unistd.h doesn't exist on Windows :pray:

Rackover:
I have it on my Windows!  Guess it's just built different 😄
The build passes regardless so I'm okay to have it in

Rackover (approved):

PR #195, opened 2025-01-10
Force include PCH

wroyca:
**What does this PR do?**

Treats the PCH as transitive (i.e. applied automatically across all compilation units) for uniform inclusion


Rackover (approved):
If I had known this existed I think i would have done it earlier, looks very useful

PR #198, opened 2025-01-16
Use {} placeholders for std::format to resolve demo metadata files not being automatically deleted.

enclave:
**What does this PR do?**

Replaces the `%s` format specifier, which is not recognized by `std::format`, with the appropriate `{}` placeholders.

I noticed that the demo metadata files (i.e., `.json`) are not being deleted whilst the demo files (i.e., `.dm_13) are.

This pull request should resolve the issue by ensuring that the metadata files residing alongside the demos are also automatically deleted based on `cl_demosKeep`.


wroyca (approved):

wroyca:
Nice catch, thank!

PR #199, opened 2025-01-16
Set the correct AIM_TARGET_INVALID value.

enclave:
**What does this PR do?**

Sets the correct `AIM_TARGET_INVALID` value.


src/Game/Structs.hpp
```diff
@@ -9656,7 +9656,7 @@ namespace Game
 		float lockOnRegionHeight;
 	};
 
-	constexpr auto AIM_TARGET_INVALID = 0x3FF;
+	constexpr auto AIM_TARGET_INVALID = 2047u; // MAX_GENTITIES - 1
```
diamante0018, line 9659: you should then MAX_GENTITIES - 1 you can move the definition from games.hpp to structs if it's not visible from there

src/Components/Modules/Bots.cpp
```diff
@@ -598,6 +603,9 @@ namespace Components
 		sv_randomBotNames = Game::Dvar_RegisterBool("sv_randomBotNames", false, Game::DVAR_NONE, "Randomize the bots' names");
 		sv_replaceBots = Game::Dvar_RegisterBool("sv_replaceBots", false, Game::DVAR_NONE, "Test clients will be replaced by connecting players when the server is full.");
 
+		AimAutoMeleeRange = Dvar::Var("aim_automelee_range");
+		PerkExtendedMeleeRange = Dvar::Var("perk_extendedMeleeRange");
+
```
diamante0018, line 608: this is the wrong way to register these dvars, they do not exist at this point in time, use dvar pointers
https://github.com/iw4x/iw4x-client/blob/develop/src/Game/Dvars.hpp#L115

src/Game/Structs.hpp
enclave: This should be rectified now. I did have to move `MAX_GENTITIES` from `Game.hpp` to `Structs.hpp`.

src/Components/Modules/Bots.cpp
enclave: As suggested, I have resorted to using Dvar pointers here.

wroyca:
Hello!

This branch was mixed with upstream so it's history is in shamble which GitHub rebase wasn't able to untangle properly, please do a interactive rebase :pray:

wroyca (dismissed):
Hello!

This branch was mixed with upstream, so its history is in shambles, and GitHub's rebase couldn't untangle it properly. Please perform an interactive rebase. 🙏

wroyca:
cc @Rackover: Not knowledgeable with the bot area, so I can only test superficially.

@enclave: It would be great to have more information here. "Sets the correct AIM_TARGET_INVALID value" tells us very little about what you're trying to solve :saluting_face:

enclave:
> cc @Rackover: Not knowledgeable with the bot area, so I can only test superficially.
> 
> @enclave: It would be great to have more information here. "Sets the correct AIM_TARGET_INVALID value" tells us very little about what you're trying to solve 🫡

Besides correcting the AIM_TARGET_INVALID constant value.

The pull request also (should) resolve the issue of bots being able to melee lunge very far! This is what I have noticed on some servers running bots. Currently the max melee charge range is 255 (`std::numeric_limits<unsigned char>::max()`), the commit sets the max melee charge range to the appropriate Dvar values (`perk_extendedMeleeRange` or `aim_automelee_range`).

src/Components/Modules/Bots.cpp
```diff
@@ -310,7 +313,9 @@ namespace Components
 			}
 
 			const auto yaw = Game::Scr_GetFloat(0);
-			const auto dist = std::clamp<int>(static_cast<int>(Game::Scr_GetFloat(1)), std::numeric_limits<unsigned char>::min(), std::numeric_limits<unsigned char>::max());
+			const auto maxDist = ent->client->ps.perks[1] & 8
```
Rackover, line 316: Might want to use PerkEnum here instead of `& 8` ?
Right now I dont see what 8 is supposed to be - `PERK_EXTENDEDMELEE`  ? If so the correct value should be 0x23 not 0x8. Are you sure this is correct and can you explain why?

src/Components/Modules/Bots.cpp
```diff
@@ -310,7 +313,9 @@ namespace Components
...
+				? static_cast<int>((*perk_extendedMeleeRange)->current.value) : static_cast<int>((*aim_automelee_range)->current.value);
+			const auto dist = std::clamp<int>(static_cast<int>(Game::Scr_GetFloat(1)), std::numeric_limits<unsigned char>::min(), maxDist);
```
Rackover, line 318: nit: There's a lot happening here, could you please unpack this on a few lines to have about an operation per line? (same thing for the one above)

```

float distanceArg = ;
const int distanceArg = static_cast<int>(Game::Scr_GetFloat(1));
const auto dist = std::clamp(distanceArg, 0, maxDist);  // (I believe the min-numeric-limit of unsigned char is just zero)
...
```

src/Components/Modules/Bots.cpp
```diff
@@ -310,7 +313,9 @@ namespace Components
 			}
 
 			const auto yaw = Game::Scr_GetFloat(0);
-			const auto dist = std::clamp<int>(static_cast<int>(Game::Scr_GetFloat(1)), std::numeric_limits<unsigned char>::min(), std::numeric_limits<unsigned char>::max());
+			const auto maxDist = ent->client->ps.perks[1] & 8
+				? static_cast<int>((*perk_extendedMeleeRange)->current.value) : static_cast<int>((*aim_automelee_range)->current.value);
```
Rackover, line 317: Nit: Check that `aim_automelee_range` exists before accessing it or early return

Rackover (changes requested):

src/Components/Modules/Bots.cpp
enclave: `8` is the correct flag here. If you x-ref one of the dvars above, disregarding the Dvar registration, there are 2 other results.

One of the x-refs is inside `AimAssist_DrawTargets` you will see the exact same check I am performing here, on which melee range Dvar is chosen.

In the other x-ref, inside `AimAssist_ApplyAutoMelee` it checks whether `isExtendedMelee` inside `AimAssistPlayerState` evaluates to true or false. Now at the very bottom of `AimAssist_BackupPlayerState`, `isExtendedMelee` is evaluated as `aaGlob->ps.isExtendedMelee = (perks[1] & 8) != 0`.

src/Components/Modules/Bots.cpp
enclave: I presume the same check would be required for `perk_extendedMeleeRange` as well. Also, to check whether the Dvar exists would that simply require checking whether the Dvar is `nullptr`?

src/Components/Modules/Bots.cpp
wroyca: Let’s give it (i.e. come up with) a name that matches, more or less, what it should be—just something approximate; should be less confusing :)

Rackover: I understand that this is valid code and that `& 8` works for your purpose, but I would like to understand (and name, if possible) this magic value before this rolls onto IW4x code. 

Logically PERK_EXTENDEDMELEE sounds consistent, but it isn't that, as you said the value doesn't match
0x8 corresponds to another perk that feels unrelated (Ninja pro). So I suppose that the playerstate perks are not from that enum, and the values mean something else.

Rackover: 0x5DCBA0 - at the bottom

Rackover:
```
    const perksEnum perkCodeIndex = BG_GetPerkCodeIndexForName(perkName); // This is a perksEnum

    const size_t  playerStatePerkSlot = perkCodeIndex >> 5; // Either 0 or 1 for our purpose

    unsigned int playerStateFlag = 1 << (perkCodeIndex & 0x1F); // <== The thing you're looking for

    entity->client->ps.perks[playerStatePerkSlot] |= playerStateFlag;
 
```

So there seems to be some clues about what calculation gives you the "player state perk flag" from the PerksEnum value (PERK_EXTENDEDMELEE).  Can you check if that calculation produces the correct numbers from you? If so, you can make a constexpr function out of it and then just use that with the provided PerksEnum to get a proper value produced by readable code, with no magic 😄   and everybody got to learn something, me first

enclave:
@Rackover I have done a bit of testing on bots and this implementation does work. Do have a review over it when you get a moment.

wroyca (approved):

PR #204, opened 2025-01-16
Remove custom console font to preserve DPI scaling

wroyca:
Using custom console font breaks DPI scaling. For example, at 200% scaling, the console window adjusts correctly, but the font size stays the same and appears extremely small. Scaling the font properly is possible, but the added complexity doesn’t justify the benefit.


Rackover:
but it looks cool

Rackover:
More seriously, the whole Console modification I made in iw4x is **purely** aesthetic. Its whole purpose is to look, kinda nice and iconic. So I'd say whatever the trouble is to bring DPI support is probably worth it - to can the feature because the font is not looking as expected on some screens, feels kind of excessive. I think this is a feature that is not purely utilitarian and rather cosmetic.

Maybe a middle-ground, in the meantime, would be to disable the font if the DPI is not set to 100%? 
If the issue is really that bad
(There's a Win32 function to check that I think)

wroyca:
Fine... but then we must bring https://int10h.org/oldschool-pc-fonts/fontlist/ collection to the table! ;)

wroyca:
> Maybe a middle-ground, in the meantime, would be to disable the font if the DPI is not set to 100%?

A middle ground would require the same amount of effort as simply fixing it, so let's address it properly at some point :beers:

PR #205, opened 2025-01-16
fix premake versioning on forks; remove unused versioning vars

mxve:
**What does this PR do?**

Fix git (and premake) related build issues on forks.
rel. #150 

So this is the cool part, it doesn't! This is the groundbreaking change that will revolutionize your gaming experience. Brace yourselves, for this PR does not alter the gameplay in any way. No, this is not some game-breaking update with new features or hidden Easter eggs. Instead, it performs the critical task of removing unused variables from the version.h file. Yes, that’s right—unused variables. These rogue pieces of code, doing absolutely nothing, are now gone. You can rest easy knowing that every line of code is now working as efficiently as possible, even if it wasn’t actually doing anything in the first place. Truly, this is the patch we've all been waiting for.

@wroyca would prefer using os.chdir in `prebuildcommands`, but I couldn't get it to work 🐒


wroyca (approved):
It works on my machine

PR #206, opened 2025-01-16
Replace repo hosted premake5 with prompted autodownload

Laupetin:
**What does this PR do?**

This PR removes the premake5 binary in the repo in favor of taking the one installed on the developers computer.
In case there is no premake5 preinstalled it offers to automatically download it.
It then places it in the build directory.

This way there is no opaque executable blob in the repository (at least for premake5) and in case of the auto downloaded version the hash can be seen very directly.
Because otherwise people would have to hash verify premake5 themselves theoretically if they wanted to check if it is the offical version (and guess the version correctly i guess).

Since it indirectly gives less control over the exact version of premake that is being used, there is a check for a minimum required premake5 version in the premake script now.

This does not impact the client, only the build process.

I did this recently in the OpenAssetTools repo and thought IW4x might benefit as well when I already brought myself to hack in this horrible scripting language 😬 If not that's also okay, this PR is just an offer for what I personally find better, it's okay if you don't :)


wroyca:
Thank!

I believe we should keep Premake in the repository. Unlike most build systems, Premake is not commonly installed system-wide, and relying on a single mirror introduces risks during downtime. Instead, my suggestion here would be to go a step further by bootstrapping Premake in the CI pipeline and replacing our version in the repository whenever there’s an upstream update

Rackover:
I'm neutral about this change.

Downloading premake5 from an outside source sounds good, but it's also an issue if you expect a clone of the repo to be a _ready and usable version_ of this repo - this changes makes it so that setting up the repo is additionally one more step than cloning it, which also requires internet.
I've been in set ups with IW4x where I would clone the repo and copy it to another computer, potentially offline, which while being a very rare use case its definitely something that can surprise somebody cloning the repo and expecting to have something already self-sufficient and ready to generate.

I'm not really up-to-date on "best practices" for this matter, so, can't really decide 😐

Laupetin:
> if you expect a clone of the repo to be a _ready and usable version_ of this repo - this changes makes it so that setting up the repo is additionally one more step than cloning it, which also requires internet. I've been in set ups with IW4x where I would clone the repo and copy it to another computer, potentially offline, which while being a very rare use case its definitely something that can surprise somebody cloning the repo and expecting to have something already self-sufficient and ready to generate.

Concerning being able to just build a clone: At least if you do a normal clone you are missing the submodules which are cloned by running generate.bat (or if you are experienced enough with git you can already use a clone command that covers this, by default it does not though).

Regarding best practices: Premake is in the situation where it is a very niche tool.
I've only ever encountered it in projects that mainly use VS.
Most C++ projects do use cmake.
Which is basically always assumed to be installed on the system.
On Windows some projects just ship with a sln file though so not all people have to install it.

Generally it is probably safe to assume that the vast majority of users (on Windows) will not have premake5 installed on their OS and added to the path since that is all a manual process and premake5 provides no guide for doing so (at least as far as i am aware).
So other than for a few power users what @wroyca said is definitely true, the auto-download would be the way almost everyone would install premake I guess.

I took a look at what other projects with premake do. I took the projects from the premake website section "Projects using premake" and skipped the ones that are not using premake anymore. Some of them don't look like they are using particularly well maintained anymore. Basically all of them use a lot more build scripting though. The first 3 are projects I personally knew so I put them first.
- FiveM: [premake5.exe in repository](https://github.com/citizenfx/fivem/blob/c95b323ade0432603f67cc73d23460c517822e28/code/tools/ci/premake5.exe)
- Hazel: Downloads premake5 automatically [using a python script](https://github.com/TheCherno/Hazel/blob/1feb70572fa87fa1c4ba784a2cfeada5b4a500db/scripts/SetupPremake.py)
- 0AD: [Downloads premake.exe](https://gitea.wildfiregames.com/0ad/0ad/src/commit/1b1e00d6c462f64011dce6e8d1ab213948eeba0d/libraries/get-windows-libs.bat#L9) from selfhosted mirror (via svn)
- Bullet3: [premake5.exe in repository](https://github.com/bulletphysics/bullet3/blob/e9c461b0ace140d5c73972760781d94b7b5eee53/build3/premake5.exe)
- Decoda: [premake4.exe in repository](https://github.com/unknownworlds/decoda/blob/09ce1c32f9cd77bbe23fd2fbef18cd93c34c8ef9/utils/premake/premake4.exe)
- MultiTheftAuto: [premake5.exe in repository](https://github.com/multitheftauto/mtasa-blue/blob/bf689fdb3ee6a9d2148e1f5499cee35d9306887b/utils/premake5.exe)
- OpenDynamicsEngine: [premake5.exe in repository](https://bitbucket.org/odedevs/ode/src/ea06e91884bc82f923e25aed23fdfe496d7a5556/build/premake4.exe?at=master)
- Recast Navigation: [User is expected to install premake5 themselves](https://github.com/recastnavigation/recastnavigation/blob/77f7e54bc8cf5a816f9f087a3e0ac391d2043be3/Docs/_2_BuildingAndIntegrating.md#building-recastdemo)
- PopHead: [premake5.exe in repo](https://github.com/SPC-Some-Polish-Coders/PopHead/blob/2bce21b1a6b3d16a2ccecf0d15faeebf6a486c81/vendor/premake/premake5.exe) though project was abandoned for a couple of years already
- Orx: Didn't really understand their build process. I think it is downloaded but not from the official source? Nonetheless they do have a couple of binaries in the repo anyway.

So IW4x does not seem to be an odd one out at least in the premake "ecosystem". Like I said, if you prefer the current way, that's okay :)  Feel free to close the PR in that case.

PR #210, opened 2025-01-20
Remove legacy unit-test code

wroyca:
Remove legacy unit-test code. Although there were intentions to use it at some point, it was never fully integrated and has since become rarely (never) used / maintained


Rackover (approved):
Yeah. In a better world we should go the other way, but right now I don't see who's going to write unit tests for the whole client and/or run them

PR #215, opened 2025-01-24
Mapvote GSC helpers

Rackover:
**What does this PR do?**
Adds various GSC functions very useful to make a map voting menu, among other things

Adds scripting functions

Not sure if LocalizeText isn't already a thing, I think it isn't, pretty sure it isn't, but maybe someone will correct me on this

- [ X] Focus on a single fix or feature; remove any unrelated formatting or code changes
- [X ] Mention any [related issues](https://github.com/iw4x/iw4x-client/issues) (put `closes #XXXX` in comment to auto-close issue when PR is merged)
- [X ] Follow our [coding conventions](https://github.com/iw4x/iw4x-client/blob/master/CODESTYLE.md)
- [X ] Minimize the number of commits

### DIFF
--- a/src/Components/Modules/Localization.cpp
+++ b/src/Components/Modules/Localization.cpp
@@ -1,5 +1,6 @@
 #include "Localization.hpp"
 #include "ArenaLength.hpp"
+#include "GSC\Script.hpp"
 
 namespace Components
 {
@@ -351,6 +352,38 @@ namespace Components
 		return mapName;
 	}
 
+	void Localization::GSCr_LocalizeText()
+	{
+		if (Game::Scr_GetNumParam() != 1)
+		{
+			Game::Scr_Error("GSCr_LocalizeText: missing key!");
+			return;
+		}
+
+		const char* str = Game::Scr_GetString(0);
+
+		const auto result = Game::UI_SafeTranslateString(str);
+
+		Game::Scr_AddString(result);
+	}
+
+	void Localization::GSCr_LocalizeGametype()
+	{
+		if (Game::Scr_GetNumParam() != 1)
+		{
+			Game::Scr_Error("GSCr_LocalizeGametype: missing gametype!");
+			return;
+		}
+
+		const char* gametype = Game::Scr_GetString(0);
+
+		// UI_GetGameTypeDisplayName
+		const auto result = Utils::Hook::Call<const char* (const char*)>(0x4EB0B0)(gametype);
+
+		Game::Scr_AddString(result);
+	}
+
+
 	Localization::Localization()
 	{
 		SetCredits();
@@ -397,6 +430,12 @@ namespace Components
 				}
 			}
 		});
+
+		// gsc: LocalizeText(key)  (e.g. MP_MIDDLE_SQUAD)
+		Components::GSC::Script::AddFunction("LocalizeText", GSCr_LocalizeText);
+
+		// gsc: LocalizeGametype(gametype)  (e.g.: dm, war, koth)
+		Components::GSC::Script::AddFunction("LocalizeGametype", GSCr_LocalizeGametype);
 	}
 
 	Localization::~Localization()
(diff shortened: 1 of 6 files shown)


src/Components/Modules/Localization.cpp
```diff
@@ -1,4 +1,5 @@
 #include "ArenaLength.hpp"
+#include "GSC\Script.hpp"
```
wroyca, line 2:
```suggestion
#include "GSC/Script.hpp"
```

src/Components/Modules/Maps.cpp
```diff
@@ -4,6 +4,7 @@
 #include "RawFiles.hpp"
 #include "StartupMessages.hpp"
 #include "Theatre.hpp"
+#include "GSC\Script.hpp"
```
wroyca, line 7:
```suggestion
#include "GSC/Script.hpp"
```

wroyca (changes requested):

wroyca (approved):

Rackover:
Wait I think the function I call by raw address should probably be registered in Game.hpp instead or functions.hpp maybe
UI_GetGameTypeName or something 0x4EB0B0

This code/feature comes from my fork, where I have much less of a standard for cleanliness so, I did not realize this

src/Components/Modules/Maps.cpp
```diff
@@ -776,6 +777,60 @@ namespace Components
 		return 0;
 	}
 
+	void Maps::GSCr_GetMapArenaInfo()
+	{
+		if (Game::Scr_GetNumParam() != 2)
+		{
+			Game::Scr_Error("GetMapArenaInfo: Needs az map name and a field name!");
```
wroyca, line 784: "az" - typo?

src/Components/Modules/Localization.cpp
```diff
@@ -350,6 +351,38 @@ namespace Components
...
+		{
+			Game::Scr_Error("GSCr_LocalizeText: missing key!");
+			return;
+		}
+
+		const char* str = Game::Scr_GetString(0);
+
+		const auto result = Game::UI_SafeTranslateString(str);
+
+		Game::Scr_AddString(result);
```
wroyca, line 366: What happens if Scr_GetString fails to retrieve a string? Can it fail to begin with, or is it safe to assume that it will always succeed?

wroyca, line 366: Maybe:

```suggestion
		const auto string = Game::UI_SafeTranslateString(Game::Scr_GetString(0));

		Game::Scr_AddString(string);
```

src/Components/Modules/Maps.cpp
```diff
@@ -776,6 +777,60 @@ namespace Components
...
+						Game::Scr_AddString(value);
+						return;
+					}
+				}
+
+				const std::string error = std::format("Could not find field {} in arena entry for map {}!\n", fieldName, mapName);
+				Game::Scr_Error(error.data());
+				return;
+			}
+		}
```
wroyca, line 815: Very little nit, but all of this seems better dealt with using predicate(s) :thinking:

src/Components/Modules/Localization.cpp
Rackover: never!

src/Components/Modules/Maps.cpp
Rackover: I'm not sure I understand what you mean

wroyca: e.g. https://en.cppreference.com/w/cpp/algorithm/find, https://en.cppreference.com/w/cpp/ranges/filter_view, or equivalent :saluting_face:

src/Components/Modules/Localization.cpp
diamante0018: This is a duplicate of https://github.com/iw4x/iw4x-client/blob/develop/src/Components/Modules/GSC/String.cpp#L148

istring can cast and localize strings that are preached by gsc. check that you are not accidentally making a dupe (code may be different but could be functionally the same)

PR #216, opened 2025-01-24
Fix/mouse clipping

SnowyWhite:
**What does this PR do?**

It ensures that the mouse cursor isn't clipped to the game window after the game has been closed.  
It's a small addition to #148, which introduced this behaviour. After quitting the game, I had to click somewhere on the desktop, otherwise I couldn't drag the mouse to another monitor (I play in fullscreen mode).

This MSDN article explains why these changes are required: 
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-clipcursor

Also... because I have severe skill issues when it comes to git, I accidentally pushed another unrelated commit.  
However, the only thing it does is to rename an engine function with its correct name.

### DIFF
--- a/src/Components/Modules/RawMouse.cpp
+++ b/src/Components/Modules/RawMouse.cpp
@@ -377,12 +377,13 @@ namespace Components
 
 	BOOL RawMouse::IN_RecenterMouse()
 	{
-		RECT clientRect;
+		RECT clientRect{ };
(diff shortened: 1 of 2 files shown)


wroyca:
Nice catch, thank!

wroyca (approved):

PR #217, opened 2025-01-26
Make include explicit

wroyca:
Clean up .cxx files by ensuring each includes its corresponding .hxx file explicitly.

### DIFF
--- a/src/Components/Loader.cpp
+++ b/src/Components/Loader.cpp
@@ -1,5 +1,7 @@
 #include <Utils/InfoString.hpp>
 
+#include "Loader.hpp"
+
 #include "Modules/ArenaLength.hpp"
 #include "Modules/Auth.hpp"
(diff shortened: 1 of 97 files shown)


Rackover (commented):
Looks good to me, okay to merge if it passes build

Rackover:
ah you closed this while i was away, sorry the tab was still open in my browser. Duplicate with #218 ?

wroyca:
> ah you closed this while i was away, sorry the tab was still open in my browser. Duplicate with #218 ?

I closed this as I don't have time to follow through and take responsibility; but feel free to cherry-pick—it builds fine. The same applies to jthread, by the way. Some minor things need more work, and I noticed a few places where std::thread is used without naming, which need replacing (and possibly some more thought about which names to assign). :saluting_face:

PR #218, opened 2025-01-26
Update CreateNamedThread with jthread

wroyca:
Let's move away from std::thread and instead use std::jthread.


Rackover:
I think this looks good (?)  not sure why it was closed.  Looked up `std::jthread` and it does sound better than `std::thread` from what I can see

PR #246, opened 2025-03-01
Net code: model index bit count update

Caball009:
**What does this PR do?**

This PR expands the list of model indices that need to be increased to handle the increased model count.

This changes the network code, so the server and client both need to be updated. Additionally, demo playback will break forward and backward compatibility due to the network code change. Maybe it's possible to check the protocol when playing back a demo and revert this change if necessary.

I presume network changes are reflected in a protocol value somewhere, which would need to be updated.

This is untested.

### DIFF
--- a/src/Components/Modules/ModelCache.cpp
+++ b/src/Components/Modules/ModelCache.cpp
@@ -35,40 +35,50 @@ namespace Components
 
 	ModelCache::ModelCache()
 	{
+		static constexpr std::array<std::string_view, 9> fieldsToUpdate
+		{
+			// clientState
+			"modelindex",
+			"attachModelIndex[0]",
+			"attachModelIndex[1]",
+			"attachModelIndex[2]",
+			"attachModelIndex[3]",
+			"attachModelIndex[4]",
+			"attachModelIndex[5]",
+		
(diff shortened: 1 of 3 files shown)


src/Components/Modules/ModelCache.cpp
```diff
@@ -35,41 +35,64 @@ namespace Components
...
+			static_assert(offsetof(Game::entityState_s, index) == 144);
+		
+			if (netfield.offset == offsetof(Game::entityState_s, index) && (netfield.name == "index"sv || netfield.name == "index.item"sv))
+			{
+				// Some of the bit lengths for these index fields are already 15
+				assert(netfield.bits <= 15);
+				
+				Utils::Hook::Set(&netfield.bits, std::max(15u, newBitLength));
+			}
+		}
```
Caball009, line 94: I'm not entirely sure whether this second netfield loop is required. If it is, it's probably easiest to set all bit counts of all 15 index netfields to a bit count of 15 (they are currently set to 10, 11 or 15).

This probably could use some more investigating and testing.

Rackover:
The protocol version you're looking for is in Structs.hpp i believe :)

PR #253, opened 2025-03-30
add LoadMod UIScript

mxve:
**What does this PR do?**

This PR adds the "LoadMod" UIScript, which is used by https://github.com/iw4x/iw4x-rawfiles/pull/40.
It allows loading a mod from a server without fully connecting to it.

Party::Connect and Download::InitiateClientDownload have been extended with the (optional) `reconnect` argument, which defaults to true. If set to false the reconnect command thats normally executed gets skipped, so the mod will be loaded but the player won't get connected to the server, instead they get placed in the main menu.

### DIFF
--- a/src/Components/Modules/ServerList.cpp
+++ b/src/Components/Modules/ServerList.cpp
@@ -1001,6 +1001,15 @@ namespace Components
 				}
 			});
 
+		UIScript::Add("LoadMod", []([[maybe_unused]] const UIScript::Token& token, [[maybe_unused]] const Game::uiInfo_s* info)
+			{
+				auto* serverInfo = GetServer(CurrentServer);
+				if (serverInfo)
+				{
+					Party::Connect(serverInfo->addr, false);
+				}
+			});
+
 		UIScript::Add("ServerSort", []([[maybe_unused]] const UIScript::Token& token, [[maybe_unused]] const Game::uiInfo_s* info)
 			{
 				const auto key = token.get<int>();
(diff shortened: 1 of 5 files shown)


mxve:
test build + menus (with hardcoded strings)
[testing.zip](https://github.com/user-attachments/files/19527405/testing.zip)

src/Components/Modules/ServerList.cpp
```diff
@@ -1001,6 +1001,15 @@ namespace Components
 				}
 			});
 
+		UIScript::Add("LoadMod", []([[maybe_unused]] const UIScript::Token& token, [[maybe_unused]] const Game::uiInfo_s* info)
```
Rackover, line 1004: I think something very similar already exists (There's `RunMod` ,  and there's another called `LoadMods` that does something completely different). 
If you're sure this functionality doesn't already exist in some form (it doesn't look like it does, for me, because this is relative to the _selected server_) please give it a more explicit name like `LoadModFromSelectedServer` or `DownloadServerMod`, even if it's a bit of a mouthful, i'd rather have it be unmistakeable

Rackover (changes requested):
Just a few things to change but i like the feature and I hope i can push it through without too much delay on my end 🙇‍♀️

src/Components/Modules/ServerList.cpp
mxve: yup, functionality seems to be different from everything that exists. renamed to DownloadServerMod

mxve:
update test build [test.zip](https://github.com/user-attachments/files/20229141/test.zip)

Rackover (approved):

PR #258, opened 2025-04-11
[chore] update zonebuilder help formatting

### LINKED ISSUE #230: iAdsTrans In/Out Time
Image

Image

Using
`zonebuilder.exe +buildmod rivals +quit`
instead of
`zonebuilder.exe -buildmod rivals -quit`
fixed it

This issue is for fixing the confusion in the console where for example this line

Image

should be a dot instead of a -

mxve:
**What does this PR do?**

image

closes #230


diamante0018:
i do wonder what kind of trickery windows does to support unnicode in c string which are utf-8 because I assume u2022 is different from just a dot like the 2E char yet this seems to work. ahah who knows.

Rackover (approved):

PR #264, opened 2025-04-29
Update new domain

mxve:
iw4x.getserve.rs -> iw4x.dev
this should be its final™ home


diamante0018 (changes requested):
please spend 30k to buy back iw4x.org domain.

Rackover (approved):

PR #267, opened 2025-05-11
add new colors

mxve:
this pr adds 15 new color codes

```
< pink
= orange
> dark purple
? teal
(@ - dummy color, seems like @ can't be typed, maybe localized string related idk)
A olive
B gray
C pure red
D pure green
E pure blue
F maroon
G hot pink
H mint
I peach
J pastel green
K light red
```
image


Rackover (approved):

Rackover:
I think we could have split the range to start at `A` instead of special characters, this would also prevent having a dud in the middle for `@`... 
But it's okay, and I don't have enough spare time to be able to support such a change actively, i don't want to block the feature, so I'll let it through

PR #271, opened 2025-06-04
Fix Lobby servers

### LINKED ISSUE #268: Lobby servers: regression in behaviour
Have a lobby server "go back to the lobby" after the game ended as it should be.

On [r4695](https://github.com/iw4x/iw4x-client/releases/tag/r4695) the lobby servers were working fine.
After https://github.com/iw4x/iw4x-client/pull/70 was added into [r4838](https://github.com/iw4x/iw4x-client/releases/tag/r4838) lobby servers were affected and once a game ended the map restarts in place without going back to the lobby, this happens in LAN with stock settings for the server and clients.

There is no workaround.

r4838 to the latest are impacted

No

diamante0018:
**What does this PR do?**

Closes #268

### DIFF
--- a/src/Components/Modules/MapRotation.cpp
+++ b/src/Components/Modules/MapRotation.cpp
@@ -217,16 +217,23 @@ namespace Components
 
 	bool MapRotation::ShouldRotate()
 	{
-		if (!Dedicated::IsEnabled() && SVDontRotate.get<bool>())
+		if (!Dedicated::IsEnabled() && Dvar::Var("party_host").get<bool>())
+		{
+			Logger::Warning(Game::CON_CHANNEL_SERVER, "Not performing map rotation as we are hosting a party!\n");
+			SVDontRotate.set(true);
+			return false;
+		}
+
+		if (Dedicated::IsEnabled() && SVDontRotate.get<bool>())
 		{
 			Logger::Print(Game::CON_CHANNEL_SERVER, "Not performing map rotation as sv_dontRotate is true\n");
-			SVDontRotate.set(false);
+			SVDontRotate.set(true);
 			return false;
 		}
 
 		if (Party::IsEnabled() && Dvar::Var("party_host").get<bool>())
 		{
-			Logger::Warning(Game::CON_CHANNEL_SERVER, "Not performing map rotation as we are hosting a party!\n");
+			Logger::Warning(Game::CON_CHANNEL_SERVER, "Not performing map rotation as we are hosting a lobby server!\n");
 			return false;
 		}
 
--- a/src/Components/Modules/MapRotation.cpp
+++ b/src/Components/Modules/MapRotation.cpp
@@ -217,17 +217,16 @@ namespace Components
 
 	bool MapRotation::ShouldRotate()
 	{
-		if (!Dedicated::IsEnabled() && Dvar::Var("party_host").get<bool>())
+		if (!Dedicated::IsEnabled() && SVDontRotate.get<bool>())
 		{
-			Logger::Warning(Game::CON_CHANNEL_SERVER, "Not performing map rotation as we are hosting a party!\n");
-			SVDontRotate.set(true);
+			Logger::Print(Game::CON_CHANNEL_SERVER, "Not performing map rotation as sv_dontRotate is true\n");
+			SVDontRotate.set(false);
 			return false;
 		}
 
-		if (Dedicated::IsEnabled() && SVDontRotate.get<bool>())
+		if (Party::IsEnabled() && Dvar::Var("party_host").get<bool>())
 		{
-			Logger::Print(Game::CON_CHANNEL_SERVER, "Not performing map rotation as sv_dontRotate is true\n");
-			SVDontRotate.set(true);
+			Logger::Warning(Game::CON_CHANNEL_SERVER, "Not performing map rotation as we are hosting a party!\n");
 			return false;
 		}
 


wroyca:
> PR #70 (merged as r4838) intentionally stopped automatic map restarts in private matches which is correct behavior for private or LAN play. Reverting that change  reintroduce the issue it was originally fixing
> 
> The issue described here is a side effect or regression, but that only means the real fix lies in restoring lobby server flow independently, not by undoing the correct private match behavior, that is, a revert would fix one flow by breaking another.
> 
> Instead of reverting, the correct course of action is to identify why lobby servers are now incorrectly treated post-match (e.g., missing state flags or session type misclassification), and then restore the proper session transition to lobby in that context only. 

 _Originally posted in [#268](https://github.com/iw4x/iw4x-client/issues/268#issuecomment-2940008280)_

mxve:
As said on Discord, won't be merging this at this time as I don't want to break/worsen the private match experience any further for lobby servers which are used rarely.

diamante0018:
@mxve @wroyca This should be what you two are asking for. The previous PR removed this check and broke lobby servers. Adding this back should do the trick:

Mind you IMO what I was 'reverting' was to a previous commit which was known to be iw4x map rotation behaviour for the entirety of the history of this client. "No one" (the community wasn't polled) asked it to be changed, I think, but it was done because someone made a PR for it. So a full revert would have been a completely valid option... Well, now at least the code should be structured to fix the issue at hand. However, this is a "patch" to fix what was broken by a PR, which arguably might not have improved the client at all. So we are back to having to test this new behaviour because we can't go back to what is *known* to be working. However, I am very happy the raw input PR was flat-out reverted. More scrutiny should go before changing IW4x. Huffman patch was heavily criticised because we gave space to someone who was enthusiastic about it, but it replaced an algorithm that was used for so many decades with a new one arguabily no one tested well enough and we had to take the word of the author that a few unit tests were actually enough to replace the OG algorithm for perfomance improvemnts that I wish were more througly mesured (since they allegedly were).  I'd just revert that one has well too be honest because while IW4x has always been a toy client for   everyone  to  learn how to mod CoD some of those  stuff should  remain  in forks and  never merged.  More like: "Great you  self taught yourself how to   make  an huffman algo  but is it  really that important to you and the iw4x community to have it merged?" Now I would also  close the "anti lag" rework PR because that's just as bad, people will  need to be  stopped from adding toy experimental code because they need to understand that they can  be as entusiastic about anything thye want  but please leave  mainline iw4x out of this because this is a really old client and touching anything can break 10 other features (which is something i was guilty of a lot in the past which is why now i'd say to hold on to these, expecially sincne we do not have *any* active gameplay testers and we have to rely on static code analysis in pull requests)

diamante0018:
This would actually be where we could actually decide to remove lobby servers entirely from iw4x and be done with it, but since current mantainers dont want to upset the few people who have used them in the entire  client  history it hasnt been done.
However no one  tests  them so they get broken and the past few years it has happened but sometimes you only discover it after  6 months so it's  not fun to deal with these issues. there is a lot of  code in iw4x that could be removed and we would only have someone complain about it in maybe 3 years because they perhaps saw a yt  video  from 2018 using xyz feature.. 😁

mxve (approved):

wroyca:
> @mxve @wroyca This should be what you two are asking for. The previous PR removed this check and broke lobby servers. Adding this back should do the trick:
> 
> Mind you IMO what I was 'reverting' was to a previous commit which was known to be iw4x map rotation behaviour for the entirety of the history of this client. "No one" (the community wasn't polled) asked it to be changed, I think, but it was done because someone made a PR for it. So a full revert would have been a completely valid option... Well, now at least the code should be structured to fix the issue at hand. However, this is a "patch" to fix what was broken by a PR, which arguably might not have improved the client at all. So we are back to having to test this new behaviour because we can't go back to what is _known_ to be working. However, I am very happy the raw input PR was flat-out reverted. More scrutiny should go before changing IW4x. Huffman patch was heavily criticised because we gave space to someone who was enthusiastic about it, but it replaced an algorithm that was used for so many decades with a new one arguabily no one tested well enough and we had to take the word of the author that a few unit tests were actually enough to replace the OG algorithm for perfomance improvemnts that I wish were more througly mesured (since they allegedly were). I'd just revert that one has well too be honest because while IW4x has always been a toy client for everyone to learn how to mod CoD some of those stuff should remain in forks and never merged. More like: "Great you self taught yourself how to make an huffman algo but is it really that important to you and the iw4x community to have it merged?" Now I would also close the "anti lag" rework PR because that's just as bad, people will need to be stopped from adding toy experimental code because they need to understand that they can be as entusiastic about anything thye want but please leave mainline iw4x out of this because this is a really old client and touching anything can break 10 other features (which is something i was guilty of a lot in the past which is why now i'd say to hold on to these, expecially sincne we do not have _any_ active gameplay testers and we have to rely on static code analysis in pull requests)


I strongly disagree with this stance, both in principle and in how it represents the goals of open source and collaborative development.

Saying a revert is valid just because the previous behavior was "known" or "historical" fundamentally misunderstands how open source evolves. The fact that something remained unchanged for a long time doesn't automatically mean it's correct/optimal and or worth preserving indefinitely. We change things in software when we identify flaws and limitations, or opportunities for improvement, not just when a "poll" is conducted.

Regarding the PRs: if a proposed change is questionable, then the proper process is code review, testing, and discussion. But reverting changes simply because "someone made a PR for it" is not a reason to roll back, it's a reason to participate more actively in the review process, to push for better validation before merging. If the change broke something, then sure, patch it. But do so based on technical reasoning, not nostalgia or discomfort with iteration.

On Huffman: yes, performance changes should be measurable. But rejecting work just because it comes from someone "enthusiastic" or "self-taught" is gatekeeping. Either the code passes review or it doesn't. Either it performs better or it doesn't. The origin of the contribution isn't what matters, the technical merit does. If we aren't benchmarking, we should be. That's not a reason to revert; that's a reason to improve our process.

As for "toy" code: IW4x has always been a modded client. That means experimentation is part of its DNA. That doesn't mean anything should be merged blindly, but it also doesn't mean nothing should ever change unless it comes from a tiny core group. If the goal is stability, then propose a testing process. Propose a long-term support branch. But don't kill innovation outright.

If you want mainline IW4x to be stable and tested, then advocate for that, with policy "and" tooling "and" CI "and" structured testing. Otherwise, discouraging contributions on the basis of "they might break things" while relying on no testers and no release criteria is contradictory. If we lack gameplay testers, let's find ways to encourage testing, not freeze development.

You're right that stability matters. But stability is a product of process, not of avoiding change.

PR #272, opened 2025-06-06
fix C2102

mxve:
msvc(?) had this surprise waiting for me, can't seem to get the project to compile anymore without the suggested changes


diamante0018:
in theory this works because the array decays to a pointer which points to 4 bytes which is read as an unsigned integer of size 4. I guess, as long as the CI is happy! and maybe this is an update with msvc that broke very old behaviour

as far as I can gather this is possibly a MSVC regression so i wish to perhaps open myself a ticket with their community support

image

diamante0018:
i guess i think how you propose your changes it makes more sense but also the other way there was no issue

PR #321, opened 2025-09-20
Release CI

Laupetin:
Add release ci workflow to build the client, create a release and upload the artifacts.

Also creates an `update.json` for the launcher.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,11 +3,9 @@ name: Build
 on:
   push:
     branches:
-      - "*"
+      - "main"
+      - "develop"
   pull_request:
-    branches:
-      - "*"
-    types: [opened, synchronize, reopened]
 
 concurrency:
   group: ${{ github.ref }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,81 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - r[0-9]+
+
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  create-release:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v5
+        - uses: ncipollo/release-action@v1
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          with:
+            draft: true
+            allowUpdates: true
+            omitBodyDuringUpdate: true
+
+  build-release:
+    name: Build release binaries
+    runs-on: windows-2022
+    steps:
+      - name: Check out files
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+          fetch-depth: 0
+
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v2
+
+      - name: Generate project files
+        run: tools/premake5 vs2022
+
+      - name: Build release binaries
+        run: msbuild /m /v:minimal /p:Configuration=Release /p:Platform=Win32 build/iw4x.sln
+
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -3,11 +3,9 @@ name: Build
 on:
   push:
     branches:
-      - "*"
+      - "main"
+      - "develop"
```
wroyca, line 7: OK 

I've renamed develop to main so we can remove the `develop` here

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+      - build-release
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: write
+    steps:
+      - uses: actions/download-artifact@v5
+      - name: Prepare release tool
+        run: |
+          wget -Orelease-tool.tar.gz https://github.com/iw4x/launcher/releases/latest/download/release-tool-x86_64-unknown-linux-gnu.tar.gz
```
wroyca, line 64:
```suggestion
          wget -O release-tool.tar.gz https://github.com/iw4x/launcher/releases/latest/download/release-tool-x86_64-unknown-linux-gnu.tar.gz
```

Typo?

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+          tar -xvf release-tool.tar.gz
+          chmod +x release-tool
+      - name: Create release information
+        run: |
+          mkdir -p ./launcher-files
+          mv -v iw4x-release/iw4x.dll launcher-files/iw4x.dll
+          ./release-tool ./launcher-files
```
wroyca, line 71: This assume executable bits, do we have them on release-tool and launcher-files?

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+      - uses: ncipollo/release-action@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          artifacts: "launcher-files/*,iw4x-release/*"
+          artifactErrorsFailBuild: true
+          allowUpdates: true
+          draft: true
+          omitBodyDuringUpdate: true
+          omitDraftDuringUpdate: true
```
wroyca, line 81:
```suggestion
          omitDraftDuringUpdate: true

```

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+
+  build-release:
+    name: Build release binaries
+    runs-on: windows-2022
+    steps:
+      - name: Check out files
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+          fetch-depth: 0
```
wroyca, line 33:
```suggestion
          fetch-depth: 1
```

We don't need to fetch everything for release, so we can set the depth to 1

wroyca (changes requested):

.github/workflows/release.yml
Laupetin: No we don't but the executable bit is set as the last command in the previous step

.github/workflows/release.yml
wroyca: > Doesn't this break when i do this?

I don't think so (famous last words), since it release on tag it should in theory always grab the tagged commit anyways

Laupetin: Okay well, the thing is that its not based on the latest tag, i was wrong on that.
It does `git rev-list --count HEAD` (`getRevisionNumber()` in premake5.lua).

So it _will_ break because it counts commits...

wroyca: ah well, lets leave it as is

wroyca (approved):

PR #342, opened 2025-11-09
Disable fullscreen on first launch

SnowyWhite:
## Summary

On first launch, the game always starts in fullscreen mode, which can cause issues on modern displays with higher resolutions than what was available at the time of the game's release.

This PR changes that behavior: On first launch (when no `players/iw4x_config.cfg` exists), the game will now start in windowed mode using a resolution automatically selected by the game. 

Subsequent launches will use `players/iw4x_config.cfg` to determine the display mode, meaning this change only applies when the config file does not yet exist.

## Additional Notes

The default value of the custom dvar `r_noborder` has been changed to false.  
This is because the game seems to always select a resolution smaller than the native screen size on first launch.  
A borderless window would look odd, and it would be impossible to move or resize it.

### DIFF
--- a/src/Components/Modules/QuickPatch.cpp
+++ b/src/Components/Modules/QuickPatch.cpp
@@ -299,16 +299,13 @@ namespace Components
 
 	void QuickPatch::CL_InitRef_Hk(Game::GfxConfiguration* config)
 	{
-		// CL_InitRef() initializes a GfxConfiguration with `defaultFullscreen` set to true.
-		// This configuration is passed to R_ConfigureRenderer(), which copies it to the global `gfxCfg` structure.
-		// In R_RegisterDvars(), r_fullscreen is registered using the value from `gfxCfg.defaultFullscreen` as default.
-
-		// By overriding `defaultFullscreen` before R_ConfigureRenderer() calls R_RegisterDvars(),
-		// we can force the game to start in windowed mode on first launch.
+		// CL_InitRef() creates a GfxConfiguration with defaultFullscreen = true,
+		// which r_fullscreen inherits as the default value. We override the field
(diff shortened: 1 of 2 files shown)


wroyca:
+1 on improving first-launch behaviour, but the mode should be fullscreen-windowed rather than a standard resizable window

src/Components/Modules/QuickPatch.cpp
```diff
@@ -297,6 +297,87 @@ namespace Components
...
+				{
+					mode_index = i;
+					break;
+				}
+			}
+		}
+
+		// Set r_mode to the resolution that matches the current monitor.
+		// Note that this dvar is registered with 2 flags: DVAR_ARCHIVE and DVAR_LATCH.
+		// To ensure the change propagates correctly, we use an API call instead of modifying the dvar directly.
```
wroyca: ditto

src/Components/Modules/QuickPatch.cpp
```diff
@@ -297,6 +297,87 @@ namespace Components
...
+		}
+
+		if (!(*Game::d3d9))
+		{
+			return;
+		}
+
+		// Get the resolution of the monitor that will be used for the game
+		HMONITOR adapterMonitor = (*Game::d3d9)->GetAdapterMonitor(adapterIndex);
+		MONITORINFO mi (sizeof (MONITORINFO));
```
wroyca:
```suggestion
		MONITORINFO mi{};
		mi.cbSize = sizeof(MONITORINFO);
```

wroyca:
Some NIT and two bug to fix otherwise LGTM

Work well on Linux as well :+1:

PR #356, opened 2026-01-29
Nuke unitialization and predestroy logic

wroyca:


Rackover:
uh....

```
>iw4x.exe -zonebuilder +quit

>echo %errorlevel%
-1073740791

```

Rackover:
at least it's unified now
```
>iw4x.exe -nosteam +quit

>echo %errorlevel%
-1073740791

```

PR #374, opened 2026-03-23
Fix memory leak in Auth::GetMachineEntropy

enclave:
The buffer allocated for `GetAdaptersInfo` wasn’t being freed properly. As the code walked the linked list, it overwrote the original pointer, so by the time `free()` was called it was operating on the traversed pointer instead of the original allocation.

This fix keeps the original pointer around and frees it after the traversal. It also adds a null check after `malloc` so we don’t pass an invalid buffer to `GetAdaptersInfo` if the allocation fails.

### DIFF
--- a/src/Components/Modules/Auth.cpp
+++ b/src/Components/Modules/Auth.cpp
@@ -544,14 +544,13 @@ namespace Components
 			if (dwResult == ERROR_BUFFER_OVERFLOW)  // This is what we're expecting
 			{
 				// Now allocate a structure of the required size.
-				auto pIpAdapterInfo = std::unique_ptr<IP_ADAPTER_INFO, decltype(&free)>(
-					reinterpret_cast<PIP_ADAPTER_INFO>(malloc(outBufLen)), free);
-				if (pIpAdapterInfo)
+				std::vector<std::uint8_t> buffer(outBufLen);
+				auto* pIpAdapterInfo = reinterpret_cast<PIP_ADAPTER_INFO>(buffer.data());
 				{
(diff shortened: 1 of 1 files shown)


wroyca (changes requested):

src/Components/Modules/Auth.cpp
```diff
@@ -544,34 +544,30 @@ namespace Components
 			if (dwResult == ERROR_BUFFER_OVERFLOW)  // This is what we're expecting
 			{
 				// Now allocate a structure of the required size.
-				PIP_ADAPTER_INFO pIpAdapterInfo = reinterpret_cast<PIP_ADAPTER_INFO>(malloc(outBufLen));
-				dwResult = GetAdaptersInfo(pIpAdapterInfo, &outBufLen);
-				if (dwResult == ERROR_SUCCESS)
+				std::vector<std::uint8_t> buffer(outBufLen);
+				auto* pIpAdapterInfo = reinterpret_cast<PIP_ADAPTER_INFO>(buffer.data());
 				{
```
Laupetin, line 549: nit: Opening a block here with another level of indentation seems unnecessary.

Laupetin (commented):
> So std::unique_ptr is one approach, but in this case we want to avoid manual allocation patterns and `std::vector<std::uint8_t>` provides exactly what we need here, that is, a RAII-backed buffer for variable-sized data

unique_ptr would have also worked via `std::make_unique<std::uint8_t[]>(outBufLen)` but std::vector also works so both are fair.

Good find @enclave

Laupetin:
Well you already merged so I guess does not matter.

wroyca:
> Well you already merged so I guess does not matter.

Sorry! :heart:

PR #377, opened 2026-04-03
Allow servers to write r_specular and r_drawSun for fps boost GSC's

fieldhack:
By removing flags from the offsets, we can allow serverside GSC's to write to the client's dvar using

```
self setClientDvar("r_specular",      1);
self setClientDvar("r_drawSun",       1);
```

which previously did not work due to their protection level.

### DIFF
--- a/src/Components/Modules/Dvar.cpp
+++ b/src/Components/Modules/Dvar.cpp
@@ -444,10 +444,10 @@ namespace Components
 		Utils::Hook::Xor<std::uint32_t>(0x4D3786, Game::DVAR_ARCHIVE);
 
 		// remove all flags for r_drawSun
-		Utils::Hook::Set<std::uint8_t>(0x51a239, Game::DVAR_NONE);
+		Utils::Hook::Xor<std::uint8_t>(0x51A239, Game::DVAR_ARCHIVE);
 
 		// remove all flags for r_specular
-		Utils::Hook::Set<std::uint8_t>(0x519c39, Game::DVAR_NONE);
+		Utils::Hook::Xor<std::uint8_t>(0x519C39, Game::DVAR_ARCHIVE);
 
 		// remove write protection from fs_game
 		Utils::Hook::Xor<std::uint32_t>(0x6431EA, Game::DVAR_INIT);
--- a/src/Components/Modules/Dvar.cpp
+++ b/src/Components/Modules/Dvar.cpp
@@ -443,10 +443,10 @@ namespace Components
 		// remove archive flags for sv_hostname
 		Utils::Hook::Xor<std::uint32_t>(0x4D3786, Game::DVAR_ARCHIVE);
 
-		// remove all flags for r_drawSun
+		// remove archive flags for r_drawSun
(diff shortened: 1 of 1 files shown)


fieldhack:
As a note - it may be worthwhile to compile a list of DVAR's that are treated as Archived.

src/Components/Modules/Dvar.cpp
```diff
@@ -443,6 +443,12 @@ namespace Components
 		// remove archive flags for sv_hostname
 		Utils::Hook::Xor<std::uint32_t>(0x4D3786, Game::DVAR_ARCHIVE);
 
+		// remove all flags for r_drawSun
+		Utils::Hook::Set<std::uint8_t>(0x51a239, Game::DVAR_NONE);
```
wroyca:
```suggestion
		Utils::Hook::Set<std::uint8_t>(0x51A239, Game::DVAR_NONE);
```

src/Components/Modules/Dvar.cpp
```diff
@@ -443,6 +443,12 @@ namespace Components
...
+
+		// remove all flags for r_specular
+		Utils::Hook::Set<std::uint8_t>(0x519c39, Game::DVAR_NONE);
```
wroyca:
```suggestion
		Utils::Hook::Set<std::uint8_t>(0x519C39, Game::DVAR_NONE);
```

wroyca:
Thank for the PR!

That approach is slightly too broad as-is. What we want here is to just relax the specific restrictions preventing modification (like the cheat/protected flags), while holding onto the flags that allow the written value to be persisted 

So basically

```
Utils::Hook::Xor<std::uint8_t>(0x000000, Game::DVAR_CHEAT | Game::DVAR_ARCHIVE);
```

fieldhack:
Adjusted the offset syntax and modified to only remove the archived flag. Neither are cheat protected currently, so I'd say not removing that flag avoids excess confusion down the line should they become so.

wroyca:
> Adjusted the offset syntax and modified to only remove the archived flag. Neither are cheat protected currently, so I'd say not removing that flag avoids excess confusion down the line should they become so.


Thanks! 

Let me chew on this for a bit. Removing the flag fixes the server issue, but it also means players' console tweaks will reset on every restart. I want to discuss this trade-off with everyone first, but I'll let you know as soon as we make a call.

wroyca:
We've talked it over and reached a consensus to keep the option saved on the user's end, so we will be closing this PR. Our main concern is that when players deal with FPS issues, their first step is to lower their own graphics settings. If we allow servers to override those local settings, it could accidentally tank the framerate for the players who rely on those. 

Regardless, thank you for putting this up, we appreciate the contribution!

Laupetin:
To add onto this: removing the archived flag means this dvar is not saved into the user config anymore. This means that setting the option in the options menu (you can set speculars on or off in the options) does not persist anymore when you restart your game and defaults to true.

Users that need the "fps boost" this gives can disable speculars themselves in the options and probably have already done so I would assume.
So in the best case this would be an inconvience having to turn it off again on every start, worst case the user does not realise it and has worse fps on many servers.

PR #382, opened 2026-04-20
gamepad: ignore LS/L3 while ADS on non-sniper weapons

awsms:
Xbox 360/PS3 Mw2 behavior: don't cancel aiming when pressing LS/L3 on a controller unless we're aiming with a sniper (to let the "steady" (hold breath) action trigger).


wroyca:
Thank for the PR!

Unfortunately it break a few too many core gameplay mechanics, so we aren't able to merge it. I'm going to close this PR, but please don't hesitate to contribute again in the future!

PR #386, opened 2026-05-03
Persist GSC SetClanTag/SetName overrides across reconnect

### LINKED ISSUE #385: Clan tag does not persist or propagate correctly (scoreboard / killfeed / map_restart)
### What happened?

While SetClanTag updates the tag in-game without reconnect, the change does not persist across map_restart. Note also that modifying clanName itself does not appear to influence the clan tag, even after reconnect.

**Steps to Reproduce**

1. Start a multiplayer session.
2. Set clan tag via GSC:

   ```c
   self SetClanTag("TAG");
   ```
3. Observe:

   * Scoreboard updates
   * Killfeed reflects new tag
4. Execute:

   ```
   map_restart
   ```
5. Observe clan tag is lost.

jaxofw:
Right now if you do `self SetClanTag("ABC")` in GSC, it works mid-session but the tag is gone the moment the player reconnects. The override map was keyed by slot and wiped on VM shutdown, so reconnects always lost it.

Switched the map to be keyed by the player's steamID and removed the VM-shutdown wipe. The override gets reapplied on every userinfo read, so listen-server reconnects keep the tag. Bots still get cleared on disconnect by slot.

Also has SetClanTag write to the local `clanName` dvar so the player's own card updates right away.

closes #385

### DIFF
--- a/src/Components/Modules/GSC/UserInfo.cpp
+++ b/src/Components/Modules/GSC/UserInfo.cpp
@@ -1,26 +1,35 @@
 #include <Utils/InfoString.hpp>
 
 #include <Components/Modules/Events.hpp>
+#include <Components/Modules/ClanTags.hpp>
 
 #include "UserInfo.hpp"
 #include "Script.hpp"
 
 namespace Components::GSC
 {
-	std::unordered_map<int, UserInfo::userInfoMap> UserInfo::UserInfoOverrides;
+	std::unordered_map<std::uint64_t, UserInfo::userInfoMap> UserInfo::UserInfoOverrides;
+
+	std::uint64_t UserInfo::GetOverrideKey(const int clientNum)
+	{
+		const auto steamID = Game::svs_clients[clientNum].steamID;
+		return steamID != 0 ? steamID : static_cast<std::uint64_t>(clientNum);
+	}
 
 	void UserInfo::SV_GetUserInfo_Stub(int index, char* buffer, int bufferSize)
 	{
 		Utils::Hook::Call<void(int, char*, int)>(0x49A160)(index, buffer, bufferSize);
 
-		Utils::InfoString map(buffer);
+		const auto it = UserInfoOverrides.find(GetOverrideKey(index));
 
-		if (!UserInfoOverrides.contains(index))
+		if (it == UserInfoOverrides.end() || it->second.empty())
 		{
-			UserInfoOverrides[index] = {};
+			return;
 		}
 
-		for (const auto& [key, val] : UserInfoOverrides[index])
+		Utils::InfoString map(buffer);
+
+		for (const auto& [key, val] : it->second)
 		{
 			if (val.empty())
 			{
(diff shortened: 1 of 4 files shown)


src/Components/Modules/GSC/UserInfo.cpp
```diff
@@ -85,7 +99,8 @@ namespace Components::GSC
 			}
 
 			Logger::Debug("Setting clanName of {} to {}", ent->s.number, clanName);
-			UserInfoOverrides[ent->s.number]["clanAbbrev"] = clanName;
+			UserInfoOverrides[GetOverrideKey(ent->s.number)]["clanAbbrev"] = clanName;
+			ClanTags::SetClanName(clanName);
```
wroyca, line 103: Possible bug?

GSC methods can be executed on any player entity (e.g., iterating through all players to set clan tags), so this may accidentally overwrite the host/local client's clan tag with the tag of the last player processed by said script

This need confirmation

src/Components/Modules/GSC/UserInfo.cpp
```diff
@@ -106,7 +122,6 @@ namespace Components::GSC
 
 		AddScriptMethods();
 
-		Events::OnVMShutdown(ClearAllOverrides);
```
wroyca, line 109: This remove the VM shutdown cleanup routine, which is wrong

src/Components/Modules/GSC/UserInfo.cpp
```diff
@@ -38,7 +47,12 @@ namespace Components::GSC
 
 	void UserInfo::ClearClientOverrides(const int clientNum)
 	{
-		UserInfoOverrides[clientNum].clear();
+		const auto steamID = Game::svs_clients[clientNum].steamID;
+
+		if (steamID == 0)
+		{
+			UserInfoOverrides.erase(static_cast<std::uint64_t>(clientNum));
+		}
```
wroyca, line 55: Possible bug?

This will only clear for bots and non-Steam players. That is, players with actual SteamIDs will skip this block

src/Components/Modules/GSC/UserInfo.cpp
```diff
@@ -38,7 +47,12 @@ namespace Components::GSC
 
 	void UserInfo::ClearClientOverrides(const int clientNum)
 	{
-		UserInfoOverrides[clientNum].clear();
+		const auto steamID = Game::svs_clients[clientNum].steamID;
```
wroyca, line 50:
```suggestion
		if (clientNum < 0 || clientNum >= MAX_CLIENTS) return;

		const auto steamID = Game::svs_clients[clientNum].steamID;
```

src/Components/Modules/GSC/UserInfo.cpp
```diff
@@ -1,26 +1,35 @@
...
 #include "Script.hpp"
 
 namespace Components::GSC
 {
-	std::unordered_map<int, UserInfo::userInfoMap> UserInfo::UserInfoOverrides;
+	std::unordered_map<std::uint64_t, UserInfo::userInfoMap> UserInfo::UserInfoOverrides;
+
+	std::uint64_t UserInfo::GetOverrideKey(const int clientNum)
+	{
+		const auto steamID = Game::svs_clients[clientNum].steamID;
```
wroyca, line 15:
```suggestion
		if (clientNum < 0 || clientNum >= MAX_CLIENTS) 
		  return 0;
		
		const auto steamID = Game::svs_clients[clientNum].steamID;
```

PR #390, opened 2026-06-12
Gsc rescale models

Santahunter:
Hey, Ive added some more core GSC functions to rescale models as you like.

Many things and mods can benefit from this.

For example some people may like to resize their warzone zone barrier with it 😉

**New Functions are:**

```
resizeModels(<xmodelName>, <factor>, [time])
resetModels(<xmodelName>, [time])
<entity> resizeModel(<factor>, [time])
<entity> resetModel([time])

```
**ResizeModel()**

resizes an object by a new factor.
` obj resizeModel(<sizeFactor>, <timeInSeconds>); // time parameter is optional, empty or 0 means instant`

e.g. 
```
obj = spawn("script_model", self.origin);
obj setModel("zw2_alduin");
obj resizeModel(3, 4); 
// Object becomes 3x as big within 4 seconds with a transition.
// You can also enter values below 1 like 0.5 to shrink the model by 50%

```

**ResetModel()**

`obj resetModel(<time>); // resets a model to its vanilla size within time. time param is also optional here.`

**ResizeModels()**

Resizes all current and future models with that name to a new size

`resizeModels(<name>, <sizeFactor>, <time>); // time optional`

e.g.
`resizeModels("zw2_alduin", 30, 5);`

**ResetModels()**

resets all current and future models with that name within given time
`resetModels("zw2_alduin", 3); // time optional, if you read it till here I think you got it `

**Reproducable GSC Code for testing:**

[scale.txt](https://github.com/user-attachments/files/28876144/scale.txt)

tinysoldier
dragon resize


wroyca:
Could we split these changes into a few smaller PRs? It's a bit too large to review effectively as one batch

Santahunter:
Hm not sure how this will work out since its just 4 gsc commands which need all the stuff mentioned to be able to also scale while an animation is played etc. I can check code again though if something is uneccessary when I get time

PR #392, opened 2026-07-18
fix: disable menu runtime state cloning

michaeloliverx:
## Summary

While investigating [OpenAssetTools issue #78](https://github.com/Laupetin/OpenAssetTools/issues/78#issuecomment-5008618232), I traced the IW4x Options menu glow to the game's fastfile menu replacement path.

`DB_DynamicCloneMenu` copies runtime state from the existing menu to its replacement by item index, assuming identical layouts. The stock `pc_options_video` has 28 items while the IW4x replacement has 52, causing unrelated items to inherit state.

This disables `DB_DynamicCloneXAssetHandler[ASSET_TYPE_MENU]`. Menus are still replaced normally without copying stale runtime state.

## Verification

- Loaded an OAT-built `iw4x_code_post_gfx_mp.ff`
- Confirmed the Options menu glow was gone
- Confirmed menu navigation still worked

The initial targeted fix and the final simplified approach are kept as separate commits for review.

### DIFF
--- a/src/Components/Modules/Menus.cpp
+++ b/src/Components/Modules/Menus.cpp
@@ -38,29 +38,6 @@ namespace Components
 
 	Utils::Memory::Allocator Menus::Allocator;
 
-	using MenuDynamicCloneHandler = int(__cdecl*)(Game::XAssetHeader, Game::XAssetHeader, Game::DB_CloneType);
-	const auto DB_DynamicCloneMenu = reinterpret_cast<MenuDynamicCloneHandler>(0x5BAD10);
-
-	// Preserve runtime menu state during replacement, but do not copy item state by index
-	// when the old and new menu layouts differ.
-	int DynamicCloneMenu(const Game::XAssetHeader newHeader, const Game::XAssetHeader existingHeader, const Game::DB_CloneType cloneType)
-	{
-		auto* newMenu = newHeader.menu;
-		auto* existingMenu = existingHeader.menu;
-		if (!newMenu || !existingMenu || newMenu->itemCount == existingMenu->itemCount)
-		{
-			return DB_DynamicCloneMenu(newHeader, existingHeader, cloneType);
-		}
-
-		if (cloneType != Game::DB_CLONE_DEFAULT && cloneType != Game::DB_CLONE_NORMAL_FROM_DEFAULT)
-		{
-			newMenu->cursorItem[0] = existingMenu->cursorItem[0];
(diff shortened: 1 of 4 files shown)


Laupetin (approved):

wroyca:
@louvenarde Sorry for the ping. I remember that, at some point, either publicly or privately, we discussed the menu system and how certain bugs had been left in place because they had become relied upon, for better or worse, and breaking backward compatibility was no-go. Would you have a chance to look at this PR with that in mind?

Laupetin:
This only removes a hook of the vanilla game that is part of DB_AddXAsset. As far as I can see, the raw menu loading of iw4x does not use that at all. Therefore this should only affect menus loaded from fastfiles.

The hook of the vanilla game looks like a convenience function of the original game devs to me.
When I looked at t5 for example, it didnt have it.

My theory is that this should not do anything meaningful anyway, as it copies runtime state from one menu to another, but there shouldnt be any circumstance where a fastfile menu is replaced by another fastfile menu while being shown anyway, right?

The only effect this vanilla code seems to have to me is to create weird side effects when trying to overwrite a menu from one fastfile with another fastfile while completely changing the contents. In the case of the linked issue it is trying to put all iw4x menus inside a fastfile instead of iwds.

Btw while the linked issue is for OAT, the issue also occurs when trying to do the same thing with iw4x Zonebuilder.

louvenarde:
I'm not familiar with menu cloning (I discovered this feature by reading this PR, that should tell you something)

The "bad behaviour we must unfortunately be compatible with" I was referring to is not specifically the menus themselves but the way they're loaded.

Normally menus are only loaded from a menu list and menu lists are explicitely loaded by the game. However... previous IW4 mods (not just IW4x) just had a habit of loading "whatever" menu files were hanging around, and replacing menu files with raw menu counterparts, never requiring a menu list / enforcing that requirement.

This behaviour is terrible but is kept for backward compatibility, although *only in game* . At the main menu, menulist loading is enforced as it should be (from my understanding of the game engine's architecture) https://github.com/iw4x/iw4x-client/blob/main/src/Components/Modules/Menus.cpp#L1336

None of this seems to be related to this PR, so I suppose I have no objection. I have looked at the code but cannot form an opinion as I don't know this system / do not understand what it was there for in the first place 🤔

wroyca:
>  I have no objection. I have looked at the code but cannot form an opinion

Likewise here, @Laupetin leaving up to you to merge :+1:

Laupetin:
@michaeloliverx Thank you very much for the fix :)

PR #395, opened 2026-07-30
feat(gsc): add `SetBrushModel` entity method

michaeloliverx:
Adds a new server-side GSC entity method:

```gsc
<entity> SetBrushModel( <index> );
```

This allows scripts to assign an existing map brush model to an entity at runtime. It is a server-side-only modification and does not require connected clients to have the newest client version.

Brush models are limited to the collision models already included in the loaded map BSP. Brush-model references in the map-ents entity string use the format `*N`, where `N` is an index into the clipmap's `cmodels` array. For example:

```text
"model" "*37"
```

For example, `mp_terminal` contains the following search-and-destroy bomb brush model:

```text
{
"script_gameobjectname" "bombzone"
"classname" "script_brushmodel"
"origin" "2081 4587 220"
"model" "*37"
}
{
"gndLt" "5c57558e01"
"ltOrigin" "2082.22 4586.62 220.123"
"classname" "script_model"
"model" "com_bomb_objective"
"angles" "0 3.18772e-006 0"
"origin" "2080.8 4586.5 190.1"
"script_exploder" "pf3167_1"
"script_gameobjectname" "bombzone"
"spawnflags" "4"
"targetname" "pf3167_auto1"
"target" "pf3167_auto2"
}
```

During map loading, these definitions are parsed and their entities are created automatically. `SetBrushModel` makes it possible to instantiate the same brush model dynamically during a match:

```gsc
brush = spawn( "script_model", origin );
brush SetBrushModel( 37 );
```

The resulting entity can be positioned, rotated, linked, and manipulated like other script entities while retaining the brush model's collision.

Brush model index `0` is a special case representing the world model, as demonstrated in the attached video. The contents of other indices depend on the map and must be inspected or tested individually.

https://github.com/user-attachments/assets/b67d99d3-46a4-465c-b8c9-c69cba5b3b93

### DIFF
--- a/src/Components/Modules/GSC/Entity.cpp
+++ b/src/Components/Modules/GSC/Entity.cpp
@@ -1,0 +1,38 @@
+#include "Entity.hpp"
+#include "Script.hpp"
+
+namespace Components::GSC
+{
+	void Entity::AddScriptMethods()
+	{
+		Script::AddMethod("SetBrushModel", [](const Game::scr_entref_t entref) // gsc: <entity> SetBrushModel(<index>)
+		{
+			if (Game::Scr_GetNumParam() != 1)
+			{
+				Game::Scr_Error("usage: <entity> SetBrushModel( <index> )\n");
+				return;
+			}
+
+			auto* ent = Game::GetEntity(entref);
+			const auto index = Game::Scr_GetInt(0);
+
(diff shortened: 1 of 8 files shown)


wroyca:
8Th8hf

PR #397, opened 2026-08-02
Enable sv_voice 1 for private matches

### LINKED ISSUE #289: sv_voice on private game
sv_voice defaults to 0, disabling it on private games unless explicitly enabled every time. especially private games should probably have voice enabled by default.

carmitrey:
Sets sv_voice 1 for private matches by default.

closes #289

### DIFF
--- a/src/Components/Modules/Voice.cpp
+++ b/src/Components/Modules/Voice.cpp
@@ -462,5 +462,16 @@ namespace Components
 
 		sv_voice = Game::Dvar_RegisterBool("sv_voice", false, Game::DVAR_NONE, "Use server side voice communications");
 		sv_alltalk = Game::Dvar_RegisterBool("sv_alltalk", false, Game::DVAR_NONE, "Allow talking across teams");
+	
+		Events::OnSVInit([]()
+		{
+			// Note that setting sv_voice here would override the value from server.cfg
+			// for dedicated servers. So leave it alone in that case.
+			//
+			if (!Dedicated::IsEnabled())
+			{
+				Command::Execute("sv_voice 1", false);
+			}
+		});
 	}
 }


src/Components/Modules/Voice.cpp
```diff
@@ -462,5 +462,14 @@ namespace Components
 
 		sv_voice = Game::Dvar_RegisterBool("sv_voice", false, Game::DVAR_NONE, "Use server side voice communications");
 		sv_alltalk = Game::Dvar_RegisterBool("sv_alltalk", false, Game::DVAR_NONE, "Allow talking across teams");
+	
+		Events::OnSVInit([]()
+		{
+			// Enable voice for private matches (listen servers)
+			if (!Dedicated::IsEnabled())
```
wroyca, line 471: That check is only really needed in certain edge cases where the dedicated server breaks. It's best to test the dedicated server without it first. If nothing breaks, we can omit it

carmitrey: I can do that and edit the PR if needed 👍

carmitrey: @wroyca so if I don't do the dedicated check at all, and if someone has `sv_voice "0"` set in `server.cfg` this change makes it to where we basically ignore that and force it to `sv_voice "1"`, even for dedicated servers. 

I only added the check because I thought there may be people who want to run a dedicated server with voice disabled and checking the server is not dedicated avoids touching dedicated servers at all.

Code tested:
```
		Events::OnSVInit([]()
		{
			Command::Execute("sv_voice 1", false);

			// Enable voice for private matches (listen servers)
			// if (!Dedicated::IsEnabled())
			// {
			// 	Command::Execute("sv_voice 1", false);
			// }
		});
```

wroyca: @carmitrey Let's leave it, that said can you update the comment to mention that for context? :)

carmitrey: For sure!

PR #398, opened 2026-08-03
Add client-side automated chat muting

### LINKED ISSUE #351: Personal chat filter (opt-in, disabled by default)
### What problem does this solve?

A personal chat filter would allow users who don't wish to see certain slurs, words, or topics communicated within the in-game chat to block those messages from being visible on their client. This would allow users to curate their social experience.

### Is your feature request related to a problem? Please describe.

I personally don't really wanna see slurs. Doesn't feel nice. I do enjoy the social aspect and shit-talking, so I don't want to disable chat entirely if I can help it.

### Describe the solution you'd like

A personal disabled by default/opt-in chat filter which uses a basic non-exhaustive list of slurs (the `iw4x_filters.json` file is likely too exhaustive for this use-case). This filter would intercept chat messages before they are shown on the client, and then remove them if they contain the matching word.

To enable this feature, a console command could be used or you could simply uncomment a line in the slurs list file. The latter would probably be easiest.

Possible extensions to this functionality would include something server side for admins to use, word replacement instead of word filtering, in-game options for toggling on/off the feature.

### Technical considerations

_No response_

### Compatibility impact

- [x] This change would be backward compatible
- [ ] This change might break compatibility with existing servers/mods
- [ ] Unsure about compatibility impact

### Describe alternatives you've considered

Turning off chat works, but then you kill the social aspect of the chat. I quite enjoy the chat, so this doesn't work well for me.
Servers on iw4x tend to operate without consideration for anyone other than nostalgic bigoted men, so trusting server admins to do the work is not a viable alternative.

### Existing workarounds

_No response_

### Supporting materials

_No response_

### Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I have limited this request to a single feature
- [x] I have provided clear use cases for this feature
- [x] I have considered the implementation complexity and compatibility impact

carmitrey:
Adds a pre-generated, user-editable file in userraw/user_filters.json that allows a player (the "victim") to immediately and silently mute other players ("offenders") from chat if a message matches a word in the filter list.

Closes #351

Implementation Notes:

- The mute only applies for the current session. If the victim leaves the server and rejoins, the mute will no longer apply unless the offender types another message containing a word in `userraw/user_filters.json`. (I assume the same holds true if the offender leaves and rejoins too although I couldn't easily test that)
- Essentially this justs hook into the same muting functionality that exists for `ESC > Mute Players`. 
- The only indication a player was muted is the mute icon on the scoreboard and a debug log message in the console
- Offender chat messages still appear in the console
- An offender can be un-muted by the victim by clicking on their name in `ESC > Mute Players` or reconnecting.
- An offender muted due to a user-blacklisted word is subsequently also voice-muted (unless the offender is on the same team) -- this is due to hooking into the pre-existing mute functionality and current limitations in team chat message parsing.
- Meets basic requirements originally noted [here](https://github.com/iw4x/iw4x-client/issues/351#issuecomment-3806362258), but has some holes that need more thought.

_Code Co-Authored by Claude via Specs written using Github issue context and Discord discussion_

### DIFF
--- a/src/Components/Modules/Chat.cpp
+++ b/src/Components/Modules/Chat.cpp
@@ -12,11 +12,15 @@ namespace Components
 	Dvar::Var Chat::cg_chatWidth;
 	Dvar::Var Chat::sv_disableChat;
 	Dvar::Var Chat::sv_sayName;
+	Dvar::Var Chat::cl_autoMuteBlacklist;
 
 	bool Chat::SendChat;
 
 	Utils::Concurrency::Container<Chat::muteList> Chat::MutedList;
 	const char* Chat::MutedListFile = "userraw/muted-users.json";
+
+	Utils::Concurrency::Container<Chat::blacklist> Chat::Blacklist;
+	const char* Chat::BlacklistFile = "userraw/user_filters.json";
 
 	bool Chat::CanAddCallback = true;
 	std::vector<Scripting::Function> Chat::SayCallbacks;
@@ -25,6 +29,13 @@ namespace Components
 	std::unique_lock<Utils::NamedMutex> Chat::Lock()
 	{
 		static Utils::NamedMutex mutex{ "iw4x-mute-list-lock" };
+		std::unique_lock lock{mutex};
+		return lock;
+	}
+
+	std::unique_lock<Utils::NamedMutex> Chat::BlacklistLock()
+	{
+		static Utils::NamedMutex mutex{ "iw4x-chat-blacklist-lock" };
 		std::unique_lock lock{mutex};
 		return lock;
 	}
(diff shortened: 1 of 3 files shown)


carmitrey:
I had some discussion questions about this PR that I proposed here before I felt this should be merged:
https://github.com/iw4x/iw4x-client/issues/351#issuecomment-5159476754

The IW4x-maintained bad word list does not yet exist outside of placeholders, but I will add it after the points in the comment above have been discussed.

---

https://github.com/iw4x/iw4x-client/issues/351#issuecomment-5159476754

@wroyca @louvenarde So I've been working on this in my fork, but I've come across a few things that I think warrant some discussion.

**Open discussions:**
- Team chat messages always include the team name in the message meaning that if the bad message comes from someone on the victim's team, the offender is never actually muted, but the message would still be hidden from chat. This means for team messages, we process every individual message from the offender rather than muting them and dropping future messages entirely. I don't think it's a huge deal to leave it this way, but thought it was worth bringing up anyways.
- The filter in it's current state can be easily evaded. For example, "s-l-u-r" or "s l u r" would not be detected by having "slur" written into the blacklist. We could include as many variations as possible by default, but leave it up to the user for anything further? We could also copy the server name filter normalization stuff, but I wanted to avoid the possibility of assuming too much of what the player actually wants filtered. I felt that if the user wants some out of pocket spelling of a word also blocked, or a loose reference to a word, it should mostly be up to them to add it (to some degree).
- Players names that contain `:` ultimately break message author detection and cannot reliably mute the offender due to how we must split `rawText` on the first occurrence of `:`, however, the message would still be ultimately hidden from the victim's chat since the body content detection isn't fully tied to author detection. Like the team chat issue, I'm not sure this is a huge issue, but thought it was worth mentioning.
- My implementation currently does not write muted users to a file so reconnecting (and maybe starting a new map) means the offender would be re-muted only after they type another message containing a blocked word. Should we write muted users to the `muted_users.json` file for it to persist? I assume this file just stores raw name strings which could be circumvented by someone changing their name, so not sure it is worth the extra effort tbh.
- Default word list is currently done in-line within the implementation of creating the file for the user if it is missing which is kind of gross and possibly harder to maintain/find. Is there a better way to handle this you all would prefer? (Although I see this is how we restrict server names, so I can keep that pattern if it is okay!)
- Is a JSON file too hard for non tech savvy users? Presumably not everyone will know how to properly format JSON, and could be prone to user error, although the code does cover handling errors if JSON is not formatted properly. I.E forgetting a comma, not quoting the strings, etc.
- Do we need to limit how many entries can be in the file? I have not tested what happens if someone does something dumb like...includes the entire Webster's dictionary. Similarly, should there also be a per-entry character limit?
- Do we want to call this a "blocklist" vs a "blacklist"? I know some people prefer no longer using the term "blacklist" and figured I'd check if this was a concern for this project.

Also the basis for the implementation adheres to the original requirements mentioned by @louvenarde -- enabled by default, done silently, etc, etc. Much of this feature already had the supporting code to function and all we needed to add was the code to generate the list and parse incoming messages. The rest of it just hooks into the flow that already existed to mute players 😄

louvenarde:
> The filter in it's current state can be easily evaded. For example, "s-l-u-r" or "s l u r" would not be detected by having "slur" written into the blacklist. We could include as many variations as possible by default, but leave it up to the user for anything further? We could also copy the server name filter normalization stuff, but I wanted to avoid the possibility of assuming too much of what the player actually wants filtered. I felt that if the user wants some out of pocket spelling of a word also blocked, or a loose reference to a word, it should mostly be up to them to add it (to some degree).

You can replace letters from the message before processing it against the blacklist. So you only put the words you want to filter in the blacklist ("slur") and then before checking the text you
- Swap 1 for i,  3 for e, 0 for o, etc
- Remove any non-alphanumeric character (including spaces)

The default list should be kept very thin imo. It's just a basic security against slurs, if you start adding e.g. swear words in it then it's gonna cause a ton of problems in non-english languages and so on, so it should be kept very small.
Typical example in french - "enough" is "assez".  "Ass" is a swear word in english so filtering swear words just messes this up. Its probably true in a lot of other languages. So whatever list you put in there, keep it small.

> My implementation currently does not write muted users to a file so reconnecting (and maybe starting a new map) means the offender would be re-muted only after they type another message containing a blocked word. Should we write muted users to the muted_users.json file for it to persist?

From reading the description of your feature, i think so yes

> Is a JSON file too hard for non tech savvy users? Presumably not everyone will know how to properly format JSON, and could be prone to user error, although the code does cover handling errors if JSON is not formatted properly. I.E forgetting a comma, not quoting the strings, etc.

If it's a list of words, JSON is not needed yes. I would go for a `.txt` where 1 word = 1 line

> Do we need to limit how many entries can be in the file? I have not tested what happens if someone does something dumb like...includes the entire Webster's dictionary. Similarly, should there also be a per-entry character limit?

Can add some boundary like max 4000 words if you want, but otherwise i'd say no, if the user puts a big file in there then a big file should be used. It must never crash the game however, so it's your call how many characters per line or per word this should have. It's not a big concern so not a lot of time needs to be spent on this decision.

I'm not very active on this project at the time, I gave my two cents because you pinged me but I have not reviewed the code so this is neither an approval nor a dismissal 📝

PR #401, opened 2026-09-01
Add server-side demo recording

NotHGM:
IW4x's existing demo recorder (Theatre) is entirely client-side: it hooks the client's own demo-write primitives and reads client-only state (Game::clientConnections, Game::cgArray) that does not exist when the same binary runs headless as a dedicated server. Server operators therefore have no way to capture demos without relying on a connecting player's machine.

This adds a ServerDemo module that records demos server-side, one file per client, in the same .dm_13 format the existing Theatre playback path already consumes.

How it works:

- Taps the per-client message the engine has already finished building, via a new Events::OnSVSendClientMessage callback fired from Voice::SV_SendClientMessages_Stub. That hook site (0x4519F5, wrapping SV_EndClientSnapshot) is already owned by Voice, so the callback is routed through Events rather than installing a second, competing hook at the same address.

- Captures the initial gamestate by hooking the call to the gamestate message builder inside SV_SendClientGameState (0x6256B3). The message is cached per client when it is sent - independent of whether a recording is active yet - and prepended as the demo's first message once recording starts. Without this a demo has no configstrings/baselines and cannot be played from frame 0.

- Forces the next snapshot to be a full, non-delta one when recording starts (client_s::header.deltaMessage = -1). Otherwise the first captured snapshot deltas against a baseline that predates the recording and is absent from the file. This is the server-side counterpart to the client recorder's clc.demowaiting keyframe wait.

- Writes type-1 client archive records alongside the network messages, carrying the recorded player's origin/velocity/viewangles/movementDir/bobCycle/ commandTime. These anchor the local viewpoint during playback; without them the demo plays but the camera has no position to attach to.

Message framing matches the engine's own CL_WriteDemoMessage:
  [msgType:1][sequence:4][payloadLen:4][reliableAcknowledge:4][huffman(body)]
Note the leading reliableAcknowledge long is stored raw, OUTSIDE the compressed body - this is why Theatre::WriteBaseline writes a bare constant there and compresses a buffer with no leading long of its own.

Bots are excluded from auto-recording. client_s::bIsTestClient cannot be consulted at connect time: SV_AddTestClient dispatches its "connect bot<n>" command - which is what fires the connect event - and only sets the flag after that returns, so a bot is indistinguishable from a human at that instant. The auto-record decision is therefore deferred by a short settle window.

Adds:
  sv_demoAutoRecord  - record every connecting client (default off)
  sv_demosKeep       - how many auto-recorded demos to retain
  serverrecord <clientnum|-1>
  serverstoprecord <clientnum|-1>

Demos are written to serverdemos/ with a .json sidecar. The sidecar includes the keys Theatre::LoadDemos requires (author, mapname, gametype, length, timestamp) - it reads them with operator[] and only catches parse_error, so a missing key escapes as an unhandled type_error and crashes the Theatre menu.

### DIFF
--- a/src/Components/Modules/ServerDemo.cpp
+++ b/src/Components/Modules/ServerDemo.cpp
@@ -1,0 +1,542 @@
+#include "ServerDemo.hpp"
+
+#include "Dedicated.hpp"
+#include "Events.hpp"
+#include "FileSystem.hpp"
+#include "Scheduler.hpp"
+
+namespace Components
+{
+	std::array<ServerDemo::Session, Game::MAX_CLIENTS> ServerDemo::Sessions;
+	std::array<std::vector<unsigned char>, Game::MAX_CLIENTS> ServerDemo::CachedGamestate;
+
+	Dvar::Var ServerDemo::SVDemoAutoRecord;
+	Dvar::Var ServerDemo::SVDemosKeep;
+
+	nlohmann::json ServerDemo::DemoInfo::to_json() const
+	{
+		std::tm tm{};
+		localtime_s(&tm, &timeStamp);
+		char dateBuf[64]{};
+		asctime_s(dateBuf, sizeof(dateBuf), &tm);
+
+		// Field names/shape intentionally match the sidecar schema already produced by the
+		// other titles' demo storage (mapname/gametype/length/date/server/timestamp), so any
+		// shared demo-browsing tooling can ingest this without special-casing IW4x.
+		return nlohmann::json
+		{
+			// These five keys are REQUIRED by Theatre::LoadDemos, which reads them with
+			// metaObject["key"].get<T>() while only catching nlohmann::json::parse_error.
+			// A missing key yields null, and .get<std::string>() on null throws type_error -
+			// which is NOT caught, so it escapes as an unhandled C++ exception and crashes the
+			// game with 0xE06D7363 the moment the Theatre menu is opened. Do not remove them.
+			{ "author", clientName },
+			{ "mapname", mapname },
+			{ "gametype", gametype },
+			{ "length", length },
+			{ "timestamp", std::to_string(timeStamp) },
+
+			// Extra keys for parity with the sidecar schema used by the other titles' demo
+			// storage, so shared demo-browsing tooling can ingest these unchanged.
+			{ "date", dateBuf },
+			{ "map", mapname },
+			{ "mod", "" },
+			{ "revision", "iw4x-serverdemo-v1" },
+			{ "server", Dvar::Var("sv_hostname").get<std::string>() },
+			{ "recordedClient", clientName },
+		};
+	}
+
+	bool ServerDemo::ValidClientNum(const int clientNum)
+	{
+		return clientNum >= 0 && clientNum < static_cast<int>(Game::MAX_CLIENTS);
+	}
+
+	std::string ServerDemo::SanitizeName(const char* name)
+	{
+		// Player names are not safe to drop into a path as-is: they carry IW colour codes
+		// ('^' followed by a digit), spaces, and potentially any byte a client felt like
+		// sending. Deliberately not using the engine's I_CleanStr here - it rewrites the
+		// string in place and we are handed a pointer straight into live client_s state.
+		std::string out;
+
+		if (!name)
+		{
+			return out;
+		}
+
+		for (auto i = 0u; name[i] != '\0' && out.size() < MAX_NAME_CHARS; ++i)
+		{
+			const auto c = static_cast<unsigned char>(name[i]);
+
+			// Skip colour code pairs entirely (^ plus the digit that follows it).
+			if (c == '^' && name[i + 1] >= '0' && name[i + 1] <= '9')
+			{
+				++i;
+				continue;
+			}
+
+			if (std::isalnum(c))
+			{
+				out.push_back(static_cast<char>(c));
+			}
+			else if (c == '_' || c == '-')
+			{
+				out.push_back(static_cast<char>(c));
+			}
+			// Everything else (spaces, punctuation, non-ASCII, control bytes) is dropped
+			// rather than substituted, to keep filenames predictable.
+		}
+
+		return out;
+	}
+
+	void ServerDemo::CleanupOldAutoDemos()
+	{
+		const auto keep = SVDemosKeep.get<int>();
+		if (keep <= 0) return;
+
+		std::vector<std::string> files;
+		auto demos = FileSystem::GetFileList(DEMO_DIR, "dm_13");
+
+		for (auto& demo : demos)
+		{
+			if (Utils::String::StartsWith(demo, AUTO_PREFIX))
+			{
+				files.push_back(demo);
+			}
+		}
+
+		// Oldest first (timestamp is embedded in the filename and sorts lexicographically
+		// the same as numerically for our fixed-width-ish naming), so trimming from the
+		// front removes the oldest entries once we're over the keep limit.
+		std::ranges::sort(files);
+
+		const auto numDel = static_cast<int>(files.size()) - keep;
+		for (auto i = 0; i < numDel; ++i)
+		{
+			Logger::Print("[ServerDemo] Deleting old auto demo {}\n", files[i]);
+			FileSystem::_DeleteFile(DEMO_DIR, files[i]);
+			FileSystem::_DeleteFile(DEMO_DIR, std::format("{}.json", files[i]));
+		}
+	}
+
+	void ServerDemo::StartRecording(const int clientNum, const bool automatic)
+	{
+		if (!ValidClientNum(clientNum)) return;
+
+		auto& session = Sessions[clientNum];
+		if (session.active)
+		{
+			Logger::Print("[ServerDemo] Client {} is already being recorded\n", clientNum);
+			return;
+		}
+
+		auto* cl = &Game::svs_clients[clientNum];
+		if (cl->header.state < Game::CS_CONNECTED)
+		{
+			Logger::Print("[ServerDemo] Client {} is not connected\n", clientNum);
+			return;
+		}
+
+		// Never record bots/test clients. On a bot-filled server this would otherwise spam
+		// one demo file per bot every map (they connect at map start, before any human), which
+		// is pure noise and would also churn through sv_demosKeep and evict real players' demos.
+		// Note: this flag is only reliable once the client has settled - see OnClientConnected.
+		if (cl->bIsTestClient)
+		{
+			if (!automatic)
+			{
+				Logger::Print("[ServerDemo] Client {} is a bot, not recording\n", clientNum);
+			}
+			return;
+		}
+
+		if (automatic)
+		{
+			CleanupOldAutoDemos();
+		}
+
+		const auto timestamp = static_cast<long long>(std::time(nullptr));
+		const std::string mapname = (*Game::sv_mapname)->current.string;
+		const std::string gametype = (*Game::sv_gametype)->current.string;
+
+		// Prefer the player's actual name over their slot number - a slot number is
+		// meaningless once the match is over. Fall back to the number only when the name
+		// sanitises down to nothing (all colour codes, or non-ASCII only).
+		auto label = SanitizeName(cl->name);
+		if (label.empty())
+		{
+			label = std::to_string(clientNum);
+		}
+
+		const auto name = std::format("{}{}_{}_{}_{}",
+			automatic ? AUTO_PREFIX : "", gametype, mapname, label, timestamp);
+		const auto path = std::format("{}{}.dm_13", DEMO_DIR, name);
+
+		const auto handle = Game::FS_FOpenFileWrite(path.data());
+		if (!handle)
+		{
+			Logger::PrintError(Game::CON_CHANNEL_ERROR, "[ServerDemo] Failed to open {} for writing\n", path);
+			return;
+		}
+
+		session = Session{};
+		session.active = true;
+		session.autoRecorded = automatic;
+		session.demoFile = handle;
+		session.messageSequence = 0;
+		session.archiveIndex = 0;
+		session.name = name;
+		session.info.mapname = mapname;
+		session.info.gametype = gametype;
+		session.info.clientName = cl->name;
+		session.info.length = Game::Sys_Milliseconds();
+		std::time(&session.info.timeStamp);
+
+		// If we captured this client's gamestate message at connect time (see
+		// OnGamestateMessage), write it as message 0 now that `session.active` is true and
+		// WriteRawMessage will actually accept it. Consume-once: a stale gamestate from an
+		// earlier connection must never attach to a later recording of the same slot.
+		auto& gamestate = CachedGamestate[clientNum];
+		if (!gamestate.empty())
+		{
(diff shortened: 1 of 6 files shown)


Caball009:
Just a couple of questions to check the level of understanding of the code and context :)

> Captures the initial gamestate by hooking the call to the gamestate message builder inside SV_SendClientGameState (0x6256B3). **The message is cached per client when it is sent - independent of whether a recording is active yet - and prepended as the demo's first message once recording starts.** Without this a demo has no configstrings/baselines and cannot be played from frame 0.

(1) So what happens if the demo starts to record after a couple of minutes? Is the gamestate (`gameState_t`) still up-to-date?

> Writes type-1 client archive records alongside the network messages, carrying the recorded player's origin/velocity/viewangles/movementDir/bobCycle/ commandTime. **These anchor the local viewpoint during playback; without them the demo plays but the camera has no position to attach to.**

(2) Why does this happen, and is there an easier fix? Please explain why this data would be desirable to store since it's already sent to clients as part of the playerstate (`playerState_t`).

> Forces the next snapshot to be a full, non-delta one when recording starts (client_s::header.deltaMessage = -1). Otherwise the first captured snapshot deltas against a baseline that predates the recording and is absent from the file. This is the server-side counterpart to the client recorder's clc.demowaiting keyframe wait.

(3) What happens if next `header.deltaMessage` exceeds 1, i.e. refers to an older snapshot prior to the non-delta snapshot?

NotHGM:
Im not an expert on this codebase, I had help from Claude working through the reverse engineering and implementation. I wanted this feature myself, know other people have wanted it too, and didnt want to sit on it, so I figured Id open the PR rather than gatekeep it. Happy to dig into any of this further or make changes if somethings off.

> 
> (1) So what happens if the demo starts to record after a couple of minutes? Is the gamestate (`gameState_t`) still up-to-date?
> 
> > Writes type-1 client archive records alongside the network messages, carrying the recorded player's origin/velocity/viewangles/movementDir/bobCycle/ commandTime. **These anchor the local viewpoint during playback; without them the demo plays but the camera has no position to attach to.**
> 

Its cached once, at connect, never refreshed. If its actually used late, its stale, and wont reflect anything that changed since. In practice it doesnt bite much auto record starts about a second after connect, so its fresh there, and a manual record on someone already connected a while just has no gamestate cached to use, so it starts mid stream instead of using bad data.

> (2) Why does this happen, and is there an easier fix? Please explain why this data would be desirable to store since it's already sent to clients as part of the playerstate (`playerState_t`).
> 
> > Forces the next snapshot to be a full, non-delta one when recording starts (client_s::header.deltaMessage = -1). Otherwise the first captured snapshot deltas against a baseline that predates the recording and is absent from the file. This is the server-side counterpart to the client recorder's clc.demowaiting keyframe wait.
> 

Its genuinely the same data, Im reading it off the same `ps` struct that gets diffed into the snapshot. Its still needed because normal gameplay renders the camera from the predicted player state, built by resimulating against local input history, not straight from the last snapshot. Thats what hides snapshot rate and latency. Demo playback has no input to predict from, so that path never runs. The stock demo player expects the view straight from these archive records instead, same as the client side recorder already does.

> (3) What happens if next `header.deltaMessage` exceeds 1, i.e. refers to an older snapshot prior to the non-delta snapshot?

Not fully handled, this is a real gap. `deltaMessage = -1` only forces the one snapshot right after record start. The engines own staleness check only fires once a delta target falls out of its own big rolling buffer, it doesnt know where my recording started. And `deltaMessage` tracks the clients ack, which lags by RTT, so for a few frames right after start a snapshot can still legitimately delta against something older than my forced keyframe and missing from the file.

Caball009:
> and a manual record on someone already connected a while **just has no gamestate cached to use**, so it starts mid stream instead of using bad data.

How is that supposed to work? You remarked that the gamestate needs to be included for a valid demo.

> The stock demo player expects the view straight from these archive records instead, same as the client side recorder already does.

I'd would like a couple of demos to verify. Please include a server side demo **with** client archive data and a server side demo **without** client archive data.

I imagine without client archive data the console is spammed with "Couldn't find exact match for servertime %i, using servertime 0" during demo playback, is that right?

>  so for a few frames right after start a snapshot can still legitimately delta against something older than my forced keyframe and missing from the file.

If this happens it pretty much corrupts the entire demo, not just for a few frames, because everything is delta updated. The CoD4X server demo implementation has this particular bug :(

NotHGM:
> > and a manual record on someone already connected a while **just has no gamestate cached to use**, so it starts mid stream instead of using bad data.
> 
> How is that supposed to work? You remarked that the gamestate needs to be included for a valid demo.
> 

Yeah that line undersold it. When theres no cached gamestate the demo isnt actually valid, its the same missing frame 0 problem I already flagged, so calling it a safe fallback was wrong, its just broken output on that path right now.

Only auto record reliably has a fresh gamestate today. The proper fix, which I havent built yet, is calling the gamestate builder directly at record start instead of relying on a cached copy from connect. That closes both the staleness issue and this missing gamestate case in one go.

> > The stock demo player expects the view straight from these archive records instead, same as the client side recorder already does.
> 
> I'd would like a couple of demos to verify. Please include a server side demo **with** client archive data and a server side demo **without** client archive data.
> 
> I imagine without client archive data the console is spammed with "Couldn't find exact match for servertime %i, using servertime 0\n" during demo playback, is that right?
> 

Dont have these yet, its nearly 10pm here so Ill get them tomorrow. Build is already up in Actions if you want to grab it sooner. Also worth knowing, Ive got a branch, server-demo-maprotate, fixing a minor issue I found testing on my own server. Havent merged it in yet since I didnt want to land another change before you've had a chance to review what's already here.

On the servertime message, sounds plausible given how the engine handles a servertime with no matching archive entry, but I havent reproduced it myself so Id rather confirm it than guess.

> > so for a few frames right after start a snapshot can still legitimately delta against something older than my forced keyframe and missing from the file.
> 
> If this happens it pretty much corrupts the entire demo, not just for a few frames, because everything is delta updated. The CoD4X server demo implementation has this particular bug :(

Yeah youre right, I had that wrong. It's not a few frames, it kills the rest of the file. Snapshot deltas off the one before it, so once one frame references something that's missing, every frame after just inherits that break, nothing resyncs on its own.

Ran into basically this exact thing earlier while building it actually, different bug but same result, one bad record and the whole rest of the file was unreadable. So yeah I buy it immediately. And good to know CoD4X has the same problem, wasn't expecting that but it lines up.

Caball009:
> The proper fix, which I havent built yet, is calling the gamestate builder directly at record start instead of relying on a cached copy from connect.

True, the correct solution would be to build the gamestate the moment it's needed and not cache it.

FWIW, that issue goes away and you can simplify a bunch of things if it's acceptable that you can only start a server demo when people connect or at the start of each map. This seems quite reasonable to me. The benefit is that you don't have to create or modify any of the game messages yourself.
You could even store all outgoing data for all clients and only record demos if auto record is enabled or if clients are added manually to the record list. That way admins can still start a server demo for a client mid-match without having to wait for the next map.

EDIT: I have created an implementation as described above for iw6; I can post a link if you want to take a look.

NotHGM:
Yeah thats a good point, honestly makes the keyframe hack pointless since gamestate and the first snapshot already line up right at that moment anyway.

The buffer everything and decide later idea is smart too, keeps admins able to add someone mid match without touching deltaMessage or worrying about corruption, since the data from connect onward was already being captured either way. Only thing Id want to think through is memory, if a match runs long and nobody enables recording for a client youre holding their whole message history the entire time, so probably wants some kind of cap or ring buffer rather than letting it grow forever. Not a blocker, just something to work out.

Also merged that fix I mentioned into this branch now, so its all in one place.

And yeah if youre happy to share the iw6 implementation link that would be great to take a look at.

Caball009:
You can just clear the buffers each map. That makes sense, anyway, because having multiple maps in a single demo is undesirable. It'll likely be a couple of megabytes per client per map (unless the match length is unexpectedly long).

> And yeah if youre happy to share the iw6 implementation link that would be great to take a look at.

[link](https://git.alterware.dev/anomaly/iw6-mod/src/branch/master/src/client/component/demo_sv_recording.cpp) There are a couple of things that are slightly different from iw4x, or things that could be implemented a bit differently like the map change detection.

This is the crucial part. There's only one hook, on `SV_Netchan_Transmit`, which captures all (compressed) game data that's about to be sent to the client. This 'drives' everything and eventually it does this:
```cpp
//
// const game::mp::client_t& client, std::span<const std::uint8_t> network_data, bool client_loading
//
auto write_data_internal = [&client, network_data, client_loading](auto& output)
{
	if (!client_loading)
	{
		// extract predicted data / client archive data from playerstate
		// and write to output (either demo buffer or demo file)
		const auto archive_index = static_cast<std::size_t>(client.header.sendMessageCount) % 256;
		sv_write_predicted_data(output, client.gentity->client->ps, archive_index);
	}

	output.write(reinterpret_cast<const char*>(network_data.data()), network_data.size());
};
```
(slightly adjusted to match the iw4x context).

src/Components/Modules/ServerDemo.cpp
```diff
@@ -0,0 +1,591 @@
...
+		// caller in the whole binary (SV_SendMessageToClient, 0x48FE90) at this call site.
+		// Every message to every client - loading or active, gamestate or snapshot or reliable
+		// command - passes through here exactly once, already fully compressed. Nothing else
+		// in this codebase hooks this address.
+		Utils::Hook(SV_NETCHAN_TRANSMIT_CALL_SITE, SV_Netchan_Transmit_Stub, HOOK_CALL).install()->quick();
+
+		Scheduler::OnGameShutdown([]
+		{
+			ResetAll(false);
+		});
```
Caball009, line 586: Alternatively, could be done in `ServerDemo`s destructor, I think.

src/Components/Modules/ServerDemo.cpp
```diff
@@ -0,0 +1,591 @@
...
+
+		constexpr unsigned char msgType = 1;
+		rec.put(msgType);
+		rec.put(data.archiveIndex);
+		rec.put(ps->origin, sizeof(float[3]));
+		rec.put(ps->velocity, sizeof(float[3]));
+		rec.put(ps->movementDir);
+		rec.put(ps->bobCycle);
+		rec.put(ps->commandTime);
+		rec.put(ps->viewangles, sizeof(float[3]));
```
Caball009, line 257: `sizeof(ps->variableName)`, 3 times

src/Components/Modules/ServerDemo.cpp
```diff
@@ -0,0 +1,591 @@
...
+			if (std::isalnum(c) || c == '_' || c == '-')
+			{
+				out.push_back(static_cast<char>(c));
+			}
+		}
+
+		return out;
+	}
+
+	bool ServerDemo::CheckMapRotation()
```
Caball009, line 52: This doesn't check if the server restarted the map. I don't expect this to work for servers that run one map 24/7. The server data holds `snapFlagServerBit` that you can check. Each restart changes the value.

src/Components/Modules/ServerDemo.cpp
```diff
@@ -0,0 +1,591 @@
...
+		}
+
+		auto& data = Clients[clientNum];
+
+		// Reconnect into the same slot: a different lastConnectTime than what we last saw means
+		// this is a genuinely new connection wearing the old slot number, not a continuation.
+		if (data.lastConnectTime && *data.lastConnectTime != cl->lastConnectTime)
+		{
+			ResetClient(clientNum, false);
+		}
```
Caball009, line 330: Is this still necessary considering the code also registers a callback on player disconnect? (`OnClientDisconnected`)

src/Components/Modules/ServerDemo.cpp
```diff
@@ -0,0 +1,591 @@
...
+
+		const auto clientLoading = (cl->header.state == Game::CS_CLIENTLOADING);
+
+		// The player's own archived viewpoint is only meaningful once they are actually
+		// active in the world - matches the loading-state skip iw6-mod uses for the same
+		// reason (there is no predicted playerState worth recording while still loading in).
+		if (!clientLoading)
+		{
+			AppendArchiveRecord(data, cl);
+		}
```
Caball009, line 349: I strongly suspect it's not necessary per se to include the client archive data if you adjust `CL_GetPredictedOriginForServerTime` ([link](https://github.com/SwagSoftware/KisakCOD/blob/32598a6aeb5426f6e2bad22a1906fdf417fc245d/src/client_mp/cl_parse_mp.cpp#L112-L117)).

src/Components/Modules/ServerDemo.cpp
```diff
@@ -0,0 +1,591 @@
...
+		// selectedLocation, and IW4x deliberately disables this field client-side anyway
+		// (Theatre::CL_WriteDemoClientArchive_Hk, "Fix issue with locationSelectionInfo").
+		constexpr int locationSelectionInfo = 0;
+		rec.put(locationSelectionInfo);
+
+		EmitRecord(data, rec.bytes());
+
+		// Rolls over at 256: the engine's own archive table has exactly that many slots and
+		// rejects anything else as a corrupt demo.
+		data.archiveIndex = (data.archiveIndex + 1) & 0xFF;
```
Caball009, line 269: `data.archiveIndex = (data.archiveIndex + 1) % std::size(clientActive_t::clientArchive)`

Caball009 (commented):
Incomplete review.

src/Components/Modules/ServerDemo.cpp
```diff
@@ -0,0 +1,591 @@
...
+	{
+		auto& data = Clients[clientNum];
+		if (data.bufferActive) return;
+
+		// Reserve up front: auto-record almost immediately opens a file and starts writing
+		// straight to disk, so the buffer never needs to hold much; a manual/admin recording
+		// might be started minutes into a match, so the buffer needs to be ready to hold a
+		// full match's worth of every message sent to this client since they connected.
+		const auto autorecord = SVDemoAutoRecord.get<bool>();
+		data.buffer.reserve(autorecord ? 8192 : (512 * 1024));
```
Caball009, line 180: This is probably a minor oversight that carried over from iw6. I just made a change there: [link](https://git.alterware.dev/anomaly/iw6-mod/commit/534e9c096bca6f2f733c14377c0da48d59652e40?style=split) (only the demo_sv_recording.cpp is relevant here).

AFAICT you also keep buffering data, even when recording a demo, so this wouldn't make sense.

src/Components/Modules/ServerDemo.cpp
```diff
@@ -0,0 +1,730 @@
...
+		StopRecording(static_cast<int>(arg), true);
+	}
+
+	ServerDemo::ServerDemo()
+	{
+		SVDemoAutoRecord = Dvar::Register<bool>("sv_demoAutoRecord", false, Game::DVAR_NONE,
+			"Automatically record a server-side demo for every connecting client (dedicated/listen server only)");
+		SVDemosKeep = Dvar::Register<int>("sv_demosKeep", 100, 1, 999, Game::DVAR_NONE,
+			"How many auto-recorded server demos to keep per rotation");
+		SVDemoMaxFileSize = Dvar::Register<int>("sv_demoMaxFileSize", 512, 0, 16384, Game::DVAR_NONE,
```
Caball009: These values seem a bit large.

PR #404, opened 2026-09-03
Fix access violation in AimAssist_ApplyAutoMelee

### LINKED ISSUE #403: Access violation in AimAssist_ApplyAutoMelee (r5072, controller subsystem backport)
Client build r5072 crashes with exception code 0xC0000005 at address 0x56A378 on match load, with no mod loaded (Mod Name: None). This is a stock client defect, not a script or mod issue.

## Root Cause

Commit f55d2874 ("Backport x64 controller subsystem to x86") bound `AimAssist_ApplyAutoMelee` as a plain stack-argument function pointer:

```cpp
AimAssist_ApplyAutoMelee_t AimAssist_ApplyAutoMelee = AimAssist_ApplyAutoMelee_t(0x56A360);
```

Its two sibling functions, added in the same commit, are correctly implemented as register-passing naked trampolines, because the underlying binary expects arguments in registers rather than on the stack:

```cpp
__declspec(naked) void AimAssist_UpdateAdsLerp(const AimInput* aimInput)
{
    __asm { mov eax, [esp+0x4]  mov ebx, 0x569AA0  call ebx  retn }
}
```

`AimAssist_ApplyAutoMelee` was left as a plain function pointer. Calling it with pushed stack arguments leaves its real register arguments holding whatever was left over from the prior call, and the function dereferences that garbage.

## Evidence

Two crash dumps show the identical fault: a WRITE to `0x569AA8`. That is `0x569AA0 + 8`, the exact address `AimAssist_UpdateAdsLerp` loads into EBX (callee-saved) one call earlier, plus 8, the offset of `AimOutput::meleeChargeYaw`. This confirms the real function reads its output pointer from EBX, which the old binding never set.

## Fix

A pull request with the fix and full write-up is linked below. Verified with a Debug and Release build (Win32, MSBuild, no warnings) and a live test on `mp_terminal`, which crashed on load before the fix and loads and runs normally after.

SVRECCO:
Fixes #403

## Root Cause

Commit f55d2874 ("Backport x64 controller subsystem to x86") bound `AimAssist_ApplyAutoMelee` as a plain stack-argument function pointer:

```cpp
AimAssist_ApplyAutoMelee_t AimAssist_ApplyAutoMelee = AimAssist_ApplyAutoMelee_t(0x56A360);
```

Its two sibling functions, added in the same commit, are correctly implemented as register-passing naked trampolines, because the underlying binary expects arguments in registers rather than on the stack:

```cpp
__declspec(naked) void AimAssist_UpdateAdsLerp(const AimInput* aimInput)
{
    __asm { mov eax, [esp+0x4]  mov ebx, 0x569AA0  call ebx  retn }
}
```

`AimAssist_ApplyAutoMelee` was left as a plain function pointer. Calling it with pushed stack arguments leaves its real register arguments holding whatever was left over from the prior call, and the function dereferences that garbage.

## Evidence

Two crash dumps show the identical fault: a WRITE to `0x569AA8`. That is `0x569AA0 + 8`, the exact address `AimAssist_UpdateAdsLerp` loads into EBX (callee-saved) one call earlier, plus 8, the offset of `AimOutput::meleeChargeYaw`. This confirms the real function reads its output pointer from EBX, which the old binding never set.

## Fix

Converted to a naked trampoline, input in EAX, output in EBX:

```cpp
__declspec(naked) void AimAssist_ApplyAutoMelee(const AimInput* input, AimOutput* output)
{
    __asm
    {
        mov eax, [esp + 0x4]
        mov ebx, [esp + 0x8]
        mov edx, 0x56A360
        call edx
        retn
    }
}
```

## Verification

Builds clean in Debug and Release (Win32, MSBuild, no warnings). Release build installed in place of the previous `iw4x.dll` and tested live: private Deathmatch on `mp_terminal`, crashed on load before the fix, loads and runs normally after.

### DIFF
--- a/src/Game/Functions.cpp
+++ b/src/Game/Functions.cpp
@@ -233,7 +233,6 @@ namespace Game
 
 	Field_Draw_t Field_Draw = Field_Draw_t(0x4F5B40);
 	Field_AdjustScroll_t Field_AdjustScroll = Field_AdjustScroll_t(0x488C10);
-	AimAssist_ApplyAutoMelee_t AimAssist_ApplyAutoMelee = AimAssist_ApplyAutoMelee_t(0x56A360);
 
 	Weapon_RocketLauncher_Fire_t Weapon_RocketLauncher_Fire = Weapon_RocketLauncher_Fire_t(0x424680);
 	Bullet_Fire_t Bullet_Fire = Bullet_Fire_t(0x4402C0);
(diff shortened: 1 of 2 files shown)


wroyca:
Thank! Unfortunately I saw the PR too late and pushed a fix already https://github.com/iw4x/iw4x-client/commit/b4e02d4cd3a4f381317260dc981d4ed91f66d4f4

PR #410, opened 2026-09-14
Fix adsTransInTime/adsTransOutTime being ignored by fastfile weapons

### LINKED ISSUE #229: iAdsTransIn/OutTime not applied correctly
change ADS speed

The iAdsTransInTime & iAdsTransOutTime doesnt change in game

changing the number to be lower or higher

jaxofw:
Fixes #229.

## The bug

`iAdsTransInTime` / `iAdsTransOutTime` are never read at runtime, so editing them in a weapon file has no effect.

`PM_UpdateAimDownSightLerp` (`0x4A8370`) advances `playerState_s::weapCommon.fWeaponPosFrac` using `BG_GetWeaponDef(...)->fOOPosAnimLength[0]` going in and `[1]` coming out — a pair of per-millisecond rates at `WeaponDef+0x590`/`+0x594`. It never touches the variant def.

Those rates are derived exactly once, in the weapon file finalize pass (`0x57AC30`, reached only from `BG_LoadWeaponCompleteDefInternal`), as `1.0f / adsTransInTime`, falling back to `1/300` and `1/500` when the time is `<= 0`. The weapon field table lines up with that: `adsTransInTime` at `+0x18` and `adsTransOutTime` at `+0x1C` are parsed, but there is a deliberate hole in the table where `fOOPosAnimLength` sits, because it is derived rather than parsed.

A weapon that comes out of a fastfile never reaches that parser, so it keeps whatever rate was baked into the zone and the two times sit in the asset as inert documentation. Raw `weapons/mp/<name>` files were always fine, which is why this only shows up on zone/mod content.

## The fix

`Weapon::LoadWeaponCompleteDef` is already the single funnel every weapon registration passes through, so the rates are re-derived there on both the raw and the zone branch, reproducing the engine's own formula including its `<= 0` fallbacks. On the raw path it is a no-op; on the fastfile path it makes the weapon file authoritative again.

Gated behind `bg_adsTransitionTimeFix` (default on, `DVAR_CODINFO` so it replicates and prediction cannot desync), matching the existing `bg_weaponOffHandFix` pattern in the same file.

The three `AssertOffset` checks in the new function are load-bearing — they pin the offsets the engine actually reads.

## Testing

Reproduced against a mod whose weapons are built into a `.ff`: 21 of 71 weapons disagreed with themselves, e.g.

| weapon | weapon file says | engine was using |
| --- | --- | --- |
| `msr_mp` | in=300 out=600 | in=400 out=600 |
| `m82_mp` | in=500 out=800 | in=400 out=600 |
| `l96a1_mp` | in=400 out=400 | in=400 out=600 |
| `lynx_mp` | in=450 out=450 | in=400 out=588 |

Builds clean in Debug and Release (Win32).


Laupetin:
This kind of is a duct-tape solution though.
It fixes a problem with bad fastfiles in the client.

Which tools are these bad fastfiles created with? The best place to fix this in would probably be Zonebuilder, no?

wroyca:
Fixed in https://github.com/iw4x/iw4x-client/pull/412

PR #412, opened 2026-09-15
Fix some minor issues from backlog

### LINKED ISSUE #124: Usermaps not starting with mp_ get read by the client but can't be loaded using commands unless in zone/english
Currently in IW4x, usermaps need to start with mp_ to be loaded through commands. Otherwise they need to be put in zone/english

Usermap files can still be read by the game but the fastfile is completely ignored by commands used to start maps

### LINKED ISSUE #229: iAdsTransIn/OutTime not applied correctly

### LINKED ISSUE #245: Add built-in GSC function tagExists
Currently the game seems to have no way to make sure the gsc scripts do not run into script runtime errors when they work with xmodels. All functions which require a tagname, e.g. 'getTagOrigin("tagname")', throw a script runtime error if the tag does not exist. But you cannot make sure it exists beforehand. 
This issue often appears on custom guns (and create lagg on every weapon change due to a new script runtime error).

Image

### LINKED ISSUE #254: ZoneBuilder crash on animations with "Rawdata found"
Sometimes Zonebuilder cannot build a zone using specific iw4xAnim files. This happens on random anims but they do work on zonetool it seems. 
The exact error message is: "Reading animation 'ai_zm_dlc2_spider_trot_f" failed, remaining raw data found!"

Examplefiles are attached. 

[testanimations.zip](https://github.com/user-attachments/files/19558189/testanimations.zip)

ai_zm_dlc2_spider_trot_f.iw4xAnim (does not work)

ai_zm_dlc2_spider_queen_idle.iw4xAnim (works)

### LINKED ISSUE #257: Clamp "setPlayerData prestige"
Players set their prestige to a high value using "setPlayerData prestige" to get a "broken" prestige, which doesn't show an icon.
They tend to use high values which result in some kind of overflow that can trigger weird issues like the host migration menu showing up.

We should probably allow setting prestige to 11 (if its safe, otherwise maybe look into how we can make it safe), but the value should be clamped before being set to players dont accidentally brick something.

### LINKED ISSUE #293: loadzone command crashes when zone doesn't exist
When using the loadzone cmd, if the zone doesn't exist the game will crash.

Could a check be added to see if the zone trying to be load is there.

I don't know if this would help but this was also an issue with a cod4x plugin that got fixed, link to the change below

https://github.com/Rex109/CoD4QOL/commit/8112518d781af63470d1205e5919a938136b76f0

### LINKED ISSUE #296: Add missing dvars
`take`
`give all`
`give ammo`
`listassetpool`

### LINKED ISSUE #324: JSON ZB weapon defs cannot override common_mp weapons
What are you trying to do?
Replace MP5K_mp base weapon def with a MP5K_mp.iw4x.json weapon def. Changes include all entires; Stats, anims, model, fx, etc.

What problem are you having?
When building a MP5K_mp.iw4x.json into a fastfile with zonebuilder, loading the mod, and calling it in game either thru give or selecting it thru the menu. The weapon file built into the game in common_mp takes priority and your mod loaded .json weapon def cannot be loaded at all unless the name doesn't match a base game weapon.

Please walk us through the steps you have taken when you encountered this problem.
I have to rename the weapon to a non base weapon name, and use give commands or custom menu functions which we don't want modders who don't have menu knowledge to have to do this much work to replace a base weapon if the weapon is more complex then just a stat change aka whole weapon change. i created a tool that takes some data from a .json and throws it data into relevant calls but .json files would be much easier to work with if we could load them on the fly to replace any loaded in common_mp.

What have you tried so far?
I've tried to load a weapon raw thru weapons/MP, double checked naming and references for name match the base weapon file. 

What version of IW4x are you using?
R4930

Anything else we should know?
i don't think so but feel free to reach out if u have anymore questions regarding the issue.

### Expected behavior

When building a mp5k_mp.iw4x.json in a specific mod and loading that mod, I expect it to replace the mp5k_mp that may be built into an already loaded file. 

### Steps to reproduce the behavior

dump common_mp, grab a base game weapon of your choice, change some stats, model, or anims, and build it into a mod. Go in game and call the weapon with give or use a menu selection to get the weapon. See that stats, model, or anim will not be updated with your new created .json weapon def but that the common_mp weapon def takes priority or .json isn't recognized fully in some way to.

### LINKED ISSUE #341: IClientEngine retrieval no longer valid due to interface change
Recent Steam updates modified the interface table and the exported CreateInterface negotiation for IClientEngine, which results in the proxy retrieving either a null pointer or an incomplete vtable. In both cases, all downstream calls that require IClientEngine (user acquisition, pipe management, ISteamUser/ISteamFriends/ISteamApps mapping) cannot be executed reliably.

### LINKED ISSUE #385: Clan tag does not persist or propagate correctly (scoreboard / killfeed / map_restart)

wroyca:
Fix #124
Fix #229
Fix #245
Fix #254
Fix #257
Fix #293
Fix #296
Fix #324
Fix #341
Fix #385


louvenarde:
What happened here? This would have not passed review

